PR #30, opened 2022-01-22
Remove rxode2Test and additional test clean-ups

### LINKED ISSUE #19: Remove rxode2Test()
As discussed in #17, we plan to remove `rxode2Test()`.  The `test` argument to `rxode2Test()` has the following values where the only values suggested in the documentation are "CRAN and not cran".

The `rxode2Test()` code suggest that it:
* runs all tests if `test="cran"`
* Runs other tests, only if `Sys.getenv("NOT_CRAN") == "true"` (I think that is set by testthat)
* And otherwise skips the tests

I think that the underlying goal is to run all tests where `test="cran"` and skip tests that aren't labeled `test="cran"` on CRAN (but run them locally and on CI).  Is that correct?

Second to that, there are several values other than "cran" for `test`.  Do they need to be preserved?  The other values used are:

* demo
* focei
* indLin
* lincmt
* lvl2
* norm
* parseLincmt
* parsing
* plot
* print
* rxuse

billdenney:
This completely removes `rxode2Test()` (fixes #19).  In doing so, it replaces all of the test code because at minimum spaces changed at the beginning of the lines.  In addition to that, it tries to standardize some of the testing idioms used.

And, it fixes some other testing issues that may be platform-specific between Linux and Windows.


billdenney:
Also fixes #30.

mattfidler:
Add withr to suggests and it will likely have a successful check

mattfidler:
Windows/mac still isn't perfect.  But it is alot better than it was.  

Thank you for your help @billdenney

PR #32, opened 2022-01-23
Fix partial argument matching

billdenney:
I always run with `options(warnPartialMatchArgs=TRUE)` which results in warnings when function argument names are only partially-matched.  So, when doing some testing, I found these two partial argument matches.


mattfidler:
Thanks.  CRAN often rejects without full option matching.  Great catch.  I will enable it for me too :smile:

PR #33, opened 2022-01-23
Documentation update (no code changes)

billdenney:
I ran devtools::document() on the current code and there was a minor doc update that was missed in a previous commit.


mattfidler:
As you work a little more you may notice that reexports is changed differently configure and after a simple `devtools::document()`

PR #44, opened 2022-01-24
Catch too little memory available, and skip the test

billdenney:
The npde test fails when there is not a lot of memory available (about 22GiB in my tests).  So, skip it if there is not that much available.  The test for "not that much available" only works on Windows, and doing the test for Linux or Mac would get complicated (https://stackoverflow.com/questions/36079906/alternative-to-rs-memory-size-in-linux).


mattfidler:
This sounds fine.  I am unsure how many people manually run these tests other than us 😄

PR #45, opened 2022-01-24
Clarify and standardize all file-related test skips

billdenney:
Some tests were skipped using `if (file.exists(...))` making the skip silent.  This switches to using `skip_if_not(file.exists(...))` throughout to make file skips clearer.  It also tries to give the file variables unique names so that a skipped test is immediately findable.

### DIFF
--- a/tests/testthat/test-cbind-study-individual.R
+++ b/tests/testthat/test-cbind-study-individual.R
@@ -36,10 +36,10 @@ if (FALSE) {
 }
 
 test_that("cbind study and individual", {
-  qs <- test_path("data-cbind-study-individual.qs")
-  skip_if_not(file.exists(qs))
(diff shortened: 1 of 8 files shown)


mattfidler:
May have to document some standard here so I (and others) can follow it.

billdenney:
The overall idea is to use this pattern:

```r
test_that("description", {
  fileSpecificToTest <- test_path("filename.qs")
  skip_if_not(file.exists(test_path(fileSpecificToTest)))
  dat <- qs::qload(fileSpecificToTest)
  # insert tests here
})
```

I agree that this should go into contributing guidelines.  (I don't think that we yet have contributing guidelines.)

mattfidler:
It should be qread I think...

billdenney:
Ah good point.  The file I was looking at used `qs::qload()`.  I'll fix that in another PR.

PR #49, opened 2022-01-25
Call rxode2 model build in separate process.

### LINKED ISSUE #43: rxUse Failure on Windows
TL;DR: Should we skip the rxUse test on Windows?

When running the `rxUse()`/`rxPkg()` tests, there is an error on Windows (ref #29).  My initial testing found that the error was here:

https://github.com/nlmixr2/rxode2/blob/2908fc9e128863178329e089dd43fc4917943905/R/rxode2.R#L1338

And the error was:

```
Error in dyn.unload(.cDllFile) : 
  shared object 'C:/Users/BILLDE~1/AppData/Local/Temp/RTMPUC~2/RXUSE-~1/rxm/inst/rx/rxm_mod.dll' was not loaded
```

I got farther when I modified the line above to be `try({dyn.unload(.cDllFile)})`, but there was a later error that I couldn't directly identify.  I'd be a bit worried to wrap the `dyn.unload()` in `try()` in a general sense.  The function seems a bit ancillary to the main purpose of `rxode2`, so I'd lean toward ignoring the error on Windows.  And possibly, we could give a warning that the function may not work on Windows for any users.

@mattfidler, What are your thoughts?

mattfidler:
Closes #43


mattfidler:
I think this should work for windows @billdenney can you check?

It is slower (because it calls a new R process for each model added to the package), but works on windows.

billdenney:
I'm testing it now, and it looks like it's working, but it also appears to be installing some packages on my system (like an updated version of RcppArmadillo, and it is reinstalling rxode2).

mattfidler:
It doesnt reinstall rxodd2 on my system.

billdenney:
The rxode2 version installed on that system was several commits old.  I think that is the reason it was to installed.

mattfidler:
Ah.  That is likely the reason

billdenney:
When I checked in the morning, the test was successful on my Windows system.

mattfidler:
Thanks Bill

PR #54, opened 2022-01-27
Allow broader tolerance for random tests due to Mac differences

billdenney:
I increased the tolerance on RNG tests so that Mac will pass, and because the precision of the tolerance is not the main goal of the test.

### DIFF
--- a/tests/testthat/test-random.R
+++ b/tests/testthat/test-random.R
@@ -399,7 +399,8 @@ test_that("rbeta tests", {
   }
   
   expect_equal(mean(f$x1), mbeta(2, 5), tolerance = 0.01)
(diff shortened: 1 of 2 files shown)


mattfidler:
Awesome, Thanks @billdenney 

Now "simply" figure out the windows issues for a successful compile again.

billdenney:
Yeah, I took a look at the windows stuff, and those failures are out of my expertise.

mattfidler:
I meant to say 

Now **we** "simply"...  Thanks again for your help

mattfidler:
I looked at them, and I skipped them.  The locking mechanism seems to have changed or something happened to have windows dlls created locked until the process is finished.  Therefore, these tests are currently skipped on windows.  This is the reason why the building of the `rxm` test model was in a separate process (#43)

PR #123, opened 2022-03-15
Better errors 120

### LINKED ISSUE #120: And a related request:  If reasonable, when there is a parsing error like this, give a clearer error message by printing out `x`?  I think that this would occur by trapping all errors from parsing and printing the most relevant part of the parse when the error occurs.
And a related request:  If reasonable, when there is a parsing error like this, give a clearer error message by printing out `x`?  I think that this would occur by trapping all errors from parsing and printing the most relevant part of the parse when the error occurs.

_Originally posted by @billdenney in https://github.com/nlmixr2/rxode2/issues/119#issuecomment-1067347532_

mattfidler:
Close #120

### DIFF
--- a/R/mu.R
+++ b/R/mu.R
@@ -911,11 +911,12 @@
   .checkForInfiniteOrNaParameters(.env)
   .checkForAtLeastOneEstimatedOrModeledParameterPerEndpoint(.env)
   if (.env$hasErrors) {
-    .errMsg <- paste0(cli::col_blue("\nmodel"), "({}) errors:\n",
+    .errMsg <- paste0(crayon::bold$blue("\nmodel"), "({}) errors:\n",
                       paste(vapply(seq_along(.env$lstExpr),
                                    function(i){
-                                     sprintf("%s\033[1m:%03d:\033[0m %s",
-                                             ifelse(is.null(.env$lstErr[[i]]), "", sprintf("\033[1m%s\033[0m\n", .env$lstErr[[i]])),
+                                     sprintf(paste0("%s", crayon::bold("%03d:"), " %s"),
+                                             ifelse(is.null(.env$lstErr[[i]]), "",
+                                                    sprintf(paste0(crayon::bold("%s"), "\n"), .env$lstErr[[i]])),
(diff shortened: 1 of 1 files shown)


mattfidler:
``` r
library(rxode2)
#> rxode2 1.0.0 using 4 threads (see ?getRxThreads)
pheno2 <- function() {
  ini({
    tcl <- log(0.008) # typical value of clearance
    tv <-  log(0.6)   # typical value of volume
    ## var(eta.cl)
    eta.cl + eta.v ~ c(1, 
                       0.01, 1) ## cov(eta.cl, eta.v), var(eta.v)
    # interindividual variability on clearance and volume
    add.err <- 0.1    # residual variability
  })
  model({
    cl <- exp(+tcl + eta.cl) # individual value of clearance
    v <- exp(tv + eta.v)    # individual value of volume
    ke <- cl / v            # elimination rate constant
    d/dt(A1) = - ke * A1    # model differential equation
    cp = A1 / v             # concentration in plasma
    cp ~ add(add.err)  # define error model
  })
}

rxode2(pheno2)
#> ℹ parameter labels from comments will be replaced by 'label()'
#> 
#> model({}) errors:
#> mu-ref err: subscript out of bounds
#> 001: cl <- exp(+tcl + eta.cl)
#> 002: v <- exp(tv + eta.v)
#> 003: ke <- cl/v
#> 004: d/dt(A1) = -ke * A1
#> 005: cp = A1/v
#> 006: cp ~ add(add.err)
#> Error: syntax/parsing errors, see above
```

<sup>Created on 2022-03-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

mattfidler:
@billdenney 

This may not be exactly what you want, but without more specific information this is likely the best that we can do...

This cannot exactly be tested easily either.  

However, it will give the line where the error occurs to help in the debugging process

billdenney:
Thanks for the fix.  I have an idea for the future about how to increase the precision of the location.  That is for another day (and yes, it's hard to test because we would have to find parse errors that aren't caught before.)

PR #129, opened 2022-03-15
simplify .plotLog

billdenney:
Simplify and test .plotLog; name .plotTime outputs (applies after #126)


billdenney:
This is just for the helper function `.plotLog()`.  It's not the plots themselves.  That's why it's more easily testable.

mattfidler:
Well the plots had 100% coverage at one time.  My guess this is still true.  So, a simple before and after should give the changes if any (if I can figure out this new interface to vdiffr)

mattfidler:
To be clear the entire `plots.R`

mattfidler:
I am unsure if it was tested or not.  I double checked it is fine.

Thanks @billdenney

PR #164, opened 2022-03-23
Quieter testing

billdenney:
This makes testing quieter with fewer messages and warnings, and more tests for messages.


billdenney:
I'm not sure what the spelling error was in the initial test run.  Should we switch spelling errors to only be a warning?  (I don't know how to do that, FYI.)

mattfidler:
The spelling is skipped on CRAN.  I don't know how to make it a warning either, though it may be in the vignette.  CRAN will not accept packages with spelling errors, and the package doesn't run aspell.  For that reason, I keep it as an error

mattfidler:
Some of the tests may be slightly different since I merged with the different method of handing ini blocks in piping.  I had to change the expected number of warnings.

I also wonder if the warning should be converted to messages to match the rest of the piping experience

billdenney:
Everything is now quieter with a merge of current main.  My guess is that users don't care much between messages and warnings.  For me, they both bring about the same amount of attention.

mattfidler:
The windows tests are having errors.  They may have to do with platform differences in handling these suppression.  It think we saw something similar in the first `testthat` types of errors we were seeing.

billdenney:
Some of the errors are related to plotting tests.  I think that is related to something odd about inclusion or exclusion of xgxr.

PR #200, opened 2022-04-29
use solve.rxUi

mattfidler:


mattfidler:
@billdenney `rxTest({})` is back.  I cannot figure out how much time it takes to test on CRAN without this.  So, please don't factor it away (or describe how to do what I am already doing in a more "compliant" way)

billdenney:
> @billdenney `rxTest({})` is back. I cannot figure out how much time it takes to test on CRAN without this. So, please don't factor it away (or describe how to do what I am already doing in a more "compliant" way)

All good. :)

Issue #207, opened 2022-05-05
Mu referencing issue


mattfidler:
May be related to:

https://github.com/nlmixr2/nlmixr2est/issues/152

The mu referencing is off:

``` r
one.cmt <- function() {
  ini({
    tka <- 0.45 ; label("Ka")
    tcl <- log(c(0, 2.7, 100)) ; label("Log Cl")
    tv <- 3.45; label("log V")
    cl.wt <- 0
    v.wt <- 0
    eta.ka ~ 0.6
    eta.cl ~ 0.3
    eta.v ~ 0.1
    add.sd <- 0.7
  })
  model({
    ka <- exp(tka + eta.ka)
    cl <- exp(tcl + eta.cl + WT * cl.wt)
    v <- exp(tv + eta.v) + WT ^ 2 * v.wt
    linCmt() ~ add(add.sd)
  })
}

ui <- rxode2::rxode(one.cmt)

ui$muRefDataFrame
#>   theta    eta level
#> 1   tka eta.ka    id
#> 2   tcl eta.cl    id

ui$muRefCurEval
#>   parameter curEval low hi
#> 1    eta.ka     exp  NA NA
#> 2       tka     exp  NA NA
#> 3    eta.cl     exp  NA NA
#> 4       tcl     exp  NA NA

ui$muRefCovariateDataFrame
#>   theta covariate covariateParameter
#> 1   tcl        WT              cl.wt
```

<sup>Created on 2022-05-05 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

mattfidler:
The `muRefCurEval` should downgrade `tcl` and `eta.cl` to additive, remove the `muRefCovariateDataFrame` `WT` perhaps...

mattfidler:
The correct place to correct this is`rxode2`.  If CRAN has more comments this needs to be fixed before submitting again.  

Otherwise, a hack should be included in the `nlmixr2est` package

PR #208, opened 2022-05-05
Test the mu referencing degradation

mattfidler:


mattfidler:
This *also* fixes mu referenced detection in nested expressions (like the odd one in #207

PR #308, opened 2022-09-29
List UI

mattfidler:


mattfidler:
Ui objects are now compressed, if the `environment` hypothesis is correct, it could possibly help with

https://github.com/nlmixr2/nlmixr2est/issues/241

PR #332, opened 2022-10-17
Add odeMethodToInt to support nlmixr2/nlmixr2est#254

billdenney:
Add odeMethodToInt to support nlmixr2/nlmixr2est#254

This will ensure that ODE method integers stay the same across packages.


mattfidler:
Thanks

mattfidler:
NOTE:  If rxode2 is accepted by CRAN this will need to be moved to nlmixr2est

billdenney:
I was thinking that it feels like a better fit here.  So, the change to nlmixr2est would just wait for the next version of rxode2.

Issue #339, opened 2022-10-17
Uninitialized subject when data mixes time = 0 and time != 0


billdenney:
In the example below, the first subject in `simfit` has no simulated results because the time 0 result is removed.  (I think that this is not an issue when the first time is not zero for both subjects.)

``` r
library(tidyverse)
library(nlmixr2)
#> Loading required package: nlmixr2data
oncology_sdm_lobo_2002 <- function() {
  description <- "Signal transduction model for delayed concentration effects on cancer cell growth"
  reference <- "Lobo ED, Balthasar JP. Pharmacodynamic modeling of chemotherapeutic effects: Application of a transit compartment model to characterize methotrexate effects in vitro. AAPS J. 2002;4(4):212-222. doi:10.1208/ps040442"
  # Values for lkng, ltau, lec50, and kmax are for methotrexate from Lobo 2002,
  # Table 2.  propErr and addErr are added as reasonable values though not from
  # Lobo 2002 where no value is apparent in the paper.
  ini({
    lkng <- log(0.02) ; label("Cell net growth rate (growth minus death) (1/hr)")
    ltau <- log(c(1, 34.1, 500)) ; label("Mean transit time of each transit compartment (hr)")
    lec50 <- log(c(1, 50, 2000)) ; label("Drug concentration reducing the cell growth by 50% (ug/mL)")
    kmax <- 0.01 ; label("Maximum drug-related reduction in cell growth (1/hr)")
    
    propErr <- c(0, 0.3) ; label("Proportional residual error (fraction)")
    addErr <- c(0, 50, 1000) ; label("Additive residual error (tumor volume units)")
  })
  model({
    # cp is the drug concentration
    kng <- exp(lkng)
    tau <- exp(ltau)
    taulast <- tau
    ec50 <- exp(lec50)
    
    edrug <- kmax*cp/(ec50 + cp)
    
    tumor(0) <- tumor0
    d/dt(transit1) <- (edrug - transit1)/tau
    d/dt(transit2) <- (transit1 - transit2)/tau
    d/dt(transit3) <- (transit2 - transit3)/tau
    d/dt(transitlast) <- transit3/tau - transitlast/taulast
    d/dt(tumor) <- kng*tumor - transitlast*tumor
    tumor ~ prop(propErr) + add(addErr)
  })
}

d_sim <-
  data.frame(
    tumor0 = 100,
    time = 0:24,
    cmt = "tumor",
[...]

PR #340, opened 2022-10-18
Try using the first observed data (nocb) for evid=9

mattfidler:


mattfidler:
@billdenney this is not the cause for #339 it is still the nlmixr resolve as originally thought.  

The only issue I see is the time varying covariates are not selected correctly.

PR #349, opened 2022-10-27
new ci

mattfidler:


billdenney:
@mattfidler, I'm thinking of making custom GitHub actions to setup each dependent package so that there is a one-liner in CI to setup all dependencies.

PR #353, opened 2022-11-04
Update CI

billdenney:
Update CI to new versions


mattfidler:
It seems that these jobs are getting killed.  They used to be able to run to completion.

mattfidler:
My guess is the rstudo guys added a kill switch to their continuous integration scripts.  According to github

https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits

It should be 6 hours.

I don't know who exactly to bother here or if we need to host our own scripts (or place some options etc)

billdenney:
I wonder if I was running too many.  GitHub also limits concurrency.  I didn't see anything on the RStudio side that suggested that they had a kill switch.

mattfidler:
With less checks, I am rerunning the release ubuntu.  Maybe the concurrency is the issue..

mattfidler:
Looks good.  Thanks

PR #356, opened 2022-11-05
Allow removal of linCmt() endpoints

### LINKED ISSUE #336: Cannot append a new ODE to a model
I'm wanting to try varying the number of transit compartments using `model()` to modify them, but I cannot.

``` r
library(tidyverse)
library(nlmixr2)
#> Loading required package: nlmixr2data
oncology_sdm_lobo_2002 <- function() {
  description <- "Signal transduction model for delayed concentration effects on cancer cell growth"
  reference <- "Lobo ED, Balthasar JP. Pharmacodynamic modeling of chemotherapeutic effects: Application of a transit compartment model to characterize methotrexate effects in vitro. AAPS J. 2002;4(4):212-222. doi:10.1208/ps040442"
  # Values for lkng, ltau, lec50, and kmax are for methotrexate from Lobo 2002,
  # Table 2.  propErr and addErr are added as reasonable values though not from
  # Lobo 2002 where no value is apparent in the paper.
  ini({
    lkng <- log(0.02) ; label("Cell net growth rate (growth minus death) (1/hr)")
    ltau <- log(c(1, 34.1, 500)) ; label("Mean transit time of each transit compartment (hr)")
    lec50 <- log(c(1, 50, 2000)) ; label("Drug concentration reducing the cell growth by 50% (ug/mL)")
    kmax <- 0.01 ; label("Maximum drug-related reduction in cell growth (1/hr)")
    
    propErr <- c(0, 0.3) ; label("Proportional residual error (fraction)")
    addErr <- c(0, 50, 1000) ; label("Additive residual error (tumor volume units)")
  })
  model({
    # cp is the drug concentration
    kng <- exp(lkng)
    tau <- exp(ltau)
    taulast <- tau
    ec50 <- exp(lec50)
    
    edrug <- kmax*cp/(ec50 + cp)
    
    tumor(0) <- tumor0
    d/dt(transit1) <- (edrug - transit1)/tau
    d/dt(transit2) <- (transit1 - transit2)/tau
    d/dt(transit3) <- (transit2 - transit3)/tau
    d/dt(transitlast) <- transit3/tau - transitlast/taulast
    d/dt(tumor) <- kng*tumor - transitlast*tumor
    tumor ~ prop(propErr) + add(addErr)
  })
}

rx_orig <- nlmixr2(oncology_sdm_lobo_2002)
rx_mod <-
  rx_orig %>%
  model(
    d/dt(transit4) <- (transit3 - transit4)/tau,
    append = TRUE
  )
#> ℹ add residual parameter `transit4` and set estimate to 1
#> Error: syntax/parsing errors:
#> the following parameter(s) were in the ini block but not in the model block: transit4
```

<sup>Created on 2022-10-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### LINKED ISSUE #355: Cannot drop `linCmt()` endpoint
``` r
library(rxode2)
#> rxode2 2.0.11 using 8 threads (see ?getRxThreads)
ui <- function() {
  ini({
    tcl <- 1
    tvc <- 1
    addSd <- 1
  })
  model({
    cl <- tcl
    vc <- tvc
    linCmt() ~ add(addSd)
  })
}
# does not work
model(ui, -linCmt()~.)
#> Error: the lhs expression '-' is not in model and cannot be modified by piping
#> the lhs expression 'linCmt()' is not in model and cannot be modified by piping

ui <- function() {
  ini({
    tcl <- 1
    tvc <- 1
    addSd <- 1
  })
  model({
    cl <- tcl
    vc <- tvc
    a <- linCmt()
    a ~ add(addSd)
  })
}
# works
model(ui, -a~.)
#> ! remove population parameter `addSd`
#>  ── rxode2-based solved PK 1-compartment model ────────────────────────────────── 
#>  ── Initalization: ──  
#> Fixed Effects ($theta): 
#> tcl tvc 
#>   1   1 
#>  ── Model (Normalized Syntax): ── 
#> function() {
#>     ini({
#>         tcl <- 1
#>         tvc <- 1
#>     })
#>     model({
#>         cl <- tcl
#>         vc <- tvc
#>         a <- linCmt()
#>     })
#> }
```

<sup>Created on 2022-11-04 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

billdenney:
Fix #355
Fix #336

This makes some additional changes that will likely be able to simplify some other parts of the code around piping.  Specifically, it add the `matchesLangTemplate()`, `.getLhs()`, and `.getRhs()` functions which can likely simplify many different parts of model modification and piping code.


billdenney:
I see an error of:

`object 'rename' not found whilst loading namespace 'rxode2'`

I think that we need to reexport rename, but I recall that there were some issues around that.  @mattfidler, can you please take a look?

mattfidler:
> I see an error of:
> 
> `object 'rename' not found whilst loading namespace 'rxode2'`
> 
> I think that we need to reexport rename, but I recall that there were some issues around that. @mattfidler, can you please take a look?

This is because you have dplyr loaded when you run `devtools::document()` and it changes `rename` from an exported function `rename.rxUi` to a `S3` method `rename, rxUi`.  You can see it in the namespace.

As a work-around, you can actually drop the `@export` tags on these two functions.  That way they are not technically exported, simply registered as dplyr is loaded.

R/piping-model.R
```diff
@@ -52,26 +52,22 @@ model.rxModelVars <- model.rxode2
...
-      if (.isTilde ||
-            identical(x[[1]], quote(`=`)) ||
-            identical(x[[1]], quote(`<-`))) {
-        .tmp <- .getVariablesFromExpression(x[[3]], ignorePipe=.isTilde)
-        .tmp <- .tmp[!(.tmp %in% .lhs)]
-        .rhs <- unique(c(.tmp, .rhs))
-        .lhs <- unique(c(.getVariablesFromExpression(x[[2]]),.lhs))
+      .isTilde <- .isEndpoint(x)
+      if (.isTilde || .isAssignment(x)) {
+        .rhs <- unique(c(.getVariablesFromExpression(.getRhs(x), ignorePipe=.isTilde), .rhs))
```
mattfidler, line 60: `.rhs` should not include expressions in `.lhs`.  This matches `rxode2`'s model variables.  You probably need to add

```r
.rhs <- .rhs[!(.rhs %in .lhs)]
```

mattfidler: This only really applies to duplicated variables in the same line, but I would code for it anyway.

R/piping-model.R
```diff
@@ -88,39 +84,59 @@ model.rxModelVars <- model.rxode2
...
+.isEndpoint <- function(expr) {
+  matchesLangTemplate(expr, str2lang(". ~ ."))
+}
+
+# Determine if the input is an assignment by being 3 long and the call part
+# being either the left arrow, right arrow, or equal sign
+.isAssignment <- function(expr) {
+  matchesLangTemplate(expr, str2lang(". <- .")) |
+    matchesLangTemplate(expr, str2lang(". = .")) |
+    matchesLangTemplate(expr, str2lang(". -> ."))
```
mattfidler, line 101: `". -> "` is not supported.  

Also use logical `or` (that is `||`).  This  should never be a vector and vectorized expressions should throw an error.

tests/testthat/test-pipeline.R
```diff
@@ -123,3 +123,111 @@ rxTest({
...
+      append = TRUE
+    )
+  expect_equal(rx_mod$state, c("tumor", "transit2"))
+  expect_equal(rx_mod$ini$est, c(2, 0.3))
+  expect_equal(rx_mod$ini$name, c("a", "propSd"))
+})
+
+# Tests of individual functions ####
+
+test_that(".getModelLineEquivalentLhsExpressionDropDdt", {
```
mattfidler, line 176: These don't work on `::check()` since you are calling non-exported functions.  I typically use the `rx$getModelLineEquivalentLhsExpressionDropDdt` to not catch the codefactor errors.

tests/testthat/test-pipeline.R
```diff
@@ -123,3 +123,111 @@ rxTest({
...
+  )
+  # drop linCmt() (issue #355)
+  expect_equal(
+    .getModelLineEquivalentLhsExpressionDropEndpoint(str2lang("-linCmt()~.")),
+    str2lang("linCmt()")
+  )
+})
+
+test_that(".getVariablesFromExpression", {
+  expect_equal(.getVariablesFromExpression(""), character())
```
mattfidler, line 214: Same comment as above, use `rx$.getVariablesFromExpression`

tests/testthat/test-pipeline.R
```diff
@@ -123,3 +123,111 @@ rxTest({
...
+  expect_equal(.getVariablesFromExpression(as.name("a")), "a")
+  expect_equal(.getVariablesFromExpression(str2lang("a~b")), c("a", "b"))
+  # only pull the state from an ODE expression
+  expect_equal(.getVariablesFromExpression(str2lang("d/dt(foo)")), "foo")
+  expect_equal(.getVariablesFromExpression(str2lang("d(foo)")), "foo")
+  expect_equal(.getVariablesFromExpression(str2lang("d(foo)|bar"), ignorePipe = TRUE), "foo")
+})
+
+test_that(".getLhs, .getRhs", {
+  expect_equal(.getLhs(str2lang("a~b")), as.name("a"))
```
mattfidler, line 225: Same as above

tests/testthat/test-utils.R
```diff
@@ -0,0 +1,10 @@
+test_that("matchesLangTemplate", {
+  expect_false(matchesLangTemplate(str2lang("d/dt(foo)"), str2lang("d/foo(.name)")))
```
mattfidler, line 2: Add a 

```r
rx <- loadNamespace("rxode2")
rx$matchesLangTemplate
```

For this file as well so it will work with `::check()`

mattfidler (changes requested):
Also reviewed the code (without running it)

R/piping-model.R
billdenney: I have that here so that it would work for most lines.  Does this accomplish what you're thinking of?

https://github.com/nlmixr2/rxode2/blob/5dcc4e0eefb5791318c77168955bd7c09b6f423a/R/piping-model.R#L63

R/piping-model.R
billdenney: I was thinking that it could eventually be vectorized, but it is not near there yet.  So, I've made it use `||`.

I removed the right arrow, too.

R/piping-model.R
mattfidler: I did see that line.  In this case, I believe it doesn't apply to duplicated variables in the line.  Defiantly an edge case.

tests/testthat/test-pipeline.R
mattfidler: For my system, when the functions are not exported, `::check()` doesn't work.  

That being said I haven't checked this branch (as mentioned above). But I have had to modify some of your pull requests later since they didn't `::check()` on my system or the CI system, so I assume it is true here. 

I will check soon.

R/piping-model.R
mattfidler: Probably need to think of  a test for this edge case, actually....

R/piping-model.R
mattfidler: Thanks.  

I think you forgot to push.

tests/testthat/test-pipeline.R
mattfidler: Well, soon takes awhile.  It seems to check now and `matchesLangTemplate()` is not exported.

To be consistent perhaps we should use `.matchesLangTemplate()` instead.

R/piping-model.R
mattfidler: I think the edge case is when you are:
- Adding a line
- Using a residual component.

Perhaps it is OK to have as is.

mattfidler: I will let you decide here.  I guess simply tell me so I am aware (that way I can resolve the conversation)

mattfidler:
Overall, I think the functions you created make the code easier to read, and therefore probably maintain.

Still, I think that you need to probably

- Align the un-exported function `matchesLangTemplate()` to start with a `.`, that way the naming convention is consistent.  (Or or you could export it if you think the function is interesting enough).
- Don't export the `rename.XXX`
- Push changes

If the `rx$` convention isn't needed any longer, perhaps we should remove all of these (and figure out what version of `testthat` where this changed).  We could also add a `pkgload::load_all()` to the tests so we don't have to worry about drifting `testthat` changes.

R/piping-model.R
billdenney: I'm not sure that I understand the edge case here.  I'd prefer to take care of it (and add a test for it, whatever is decided) rather than leave it ambiguous.

As I see those lines, the overall goal is to add new `ini()` parts to the model, if necessary.  And, it does so by:

1. Accumulate all new left-hand-side (LHS) variables and right-hand-side (RHS) variables.
2. Anything on the RHS would be added to the `ini()` if it is not on the LHS.

An example model that should have an issue would be something like the following:

```
uiFun <- function() {
  ini({
    a <- 1
  })
  model({
    b <- a + c
    c <- 2
  })
}
```

The problem is that `c` is not defined when `b` asks for it, but then it is later defined.  I think that models like this should be an error regardless of where they come from (model modification or an original model).  The error would be that in the `b <- a + c` line, `c` must come from the dataset since it is not in `ini()`, and on the `c <- 2` line, information from the dataset would be overwritten.  Is this what you were thinking of?

billdenney:
As for `.matchesLangTemplate()` being exported, I do think that it is a useful function, but exporting it from `rxode2` does not seem like the right fit.  I can imagine using it in `nlmixr2lib` for some of the model modification that happens there, but maybe the right long-term solution is to consolidate all the model modification stuff in one place in which case some of the `nlmixr2lib` functions should probably migrate over here (and be exported).

R/piping-model.R
mattfidler: No, the above case should work fine.

I was thinking of the case

```
A <- A+eta
```

This occurs rarely but I think is still supported in nlmixr and rxode2.  There are even tests in rxode2 for this kind of behavior.

However this is still captures by the model variables except when they are on endpoint lines. 

This is where it could break since they arent included in the model variables

Something like

```r
Err ~add(err)
```

Which may still work ok...

mattfidler: Besides piping comparible models do not need to be compatible with whatever runs I guess.

For now I will mark it as teaolved because the edge case may not even exist

PR #357, opened 2022-11-06
Documentation updates (no code changes)

billdenney:


mattfidler:
This is a byproduct of #356 .  Since there is a bit more to do there, can you simply document on that pull request?

mattfidler:
It may change since you are moving the S3 method out anyway.

billdenney:
Yes, I'll remove the `@export` there.

PR #362, opened 2022-11-06
Remove loadNamespace() wherever possible

### LINKED ISSUE #359: Simplify and speed testing
> If the `rx$` convention isn't needed any longer, perhaps we should remove all of these (and figure out what version of `testthat` where this changed).  We could also add a `pkgload::load_all()` to the tests so we don't have to worry about drifting `testthat` changes.

_Originally posted by @mattfidler in https://github.com/nlmixr2/rxode2/issues/356#issuecomment-1304872855_

billdenney:
Fix #359

I tried to make these changes simple without making almost any other changes.  It ran successfully on my system.  @mattfidler, it's probably best to confirm that it doesn't cause problems on your system before merging it.


mattfidler:
It seems to be fine on github.  It stopped on m computer, but I think I was running it concurrently with a`testthat` so let me run it independently.

mattfidler:
Seems fine on rerun.

PR #368, opened 2022-11-29
Don't show warnings for directory creation

billdenney:
I saw an unnecessary warning for directory creation, so I removed all dir.create() warnings.  I think that this just makes it quieter without changing functionality.


mattfidler:
Looks good

PR #380, opened 2022-12-13
Make testing quieter

billdenney:
This mostly makes testing quieter.  It also has some simplifications (like changing structure(list(...)) into data.frame()).

The only serious change is that it removes the et testing that is already (confirmed) in rxode2et.


mattfidler:
I believe that et has a stub in rxode.  Perhaps keep one test to increase coverage.

billdenney:
The testing last night had errors building the models.  Hopefully the merge of main in here will fix that.

billdenney:
For the note about `et()`, I see some stuff for `etTrans` in rxode2, but I don't see any `et` code.  It appears just to be a reexport.  I did a diff of the testing file, and after adjusting white spaces, there was no difference.  If CI shows a decrease in code coverage, I'll track it down.

billdenney:
The lines of coverage are identical with this PR compared to main:  https://app.codecov.io/gh/nlmixr2/rxode2/tree/test-simplify

If you see something (admittedly hard given the size of the PR) that you'd like changed, please let me know.

PR #384, opened 2022-12-22
Allow modification of labels with ini() piping

### LINKED ISSUE #351: Feature request: allow labels in model piping
When adding a new parameter with model piping, the parameter label will be empty.  I think it would be helpful to also be able to add a `label()`.

Is there already a way to do that?

billdenney:
Fix #351 

There are some more changes that were code simplifications that I made along the way.  You can view them separately with the earlier commits.

This also gives the pointer to use label() if a character string is given.


billdenney:
This is incorporated into #385

PR #385, opened 2022-12-22
Allow reordering parameters

### LINKED ISSUE #352: Feature request: allow insertion of a parameter after a specified parameter in the ini block
When piping in new parameters, they are (mostly) inserted at the end of the parameter list.  The order of parameters can make the reporting table easier to interpret.

I think it would be helpful to have an ordering parameter to the code to pipe a new parameter in the ini block.  Does that seem useful?

billdenney:
Fix #352 

This is on top of #351 because some of the changes there improve this PR (and otherwise, there would have been many merge conflicts).

This allows modifying parameter order in model piping with `ini(param <- after(param2))` where if `param2` is not given (`after()`) it goes to the top, or you can say `after(Inf)` to go to the bottom.  (To go to the bottom, you can use any number at or higher than the maximum number of parameters in your model.)


mattfidler:
I think the `ini(parameter=label("label name"))` makes sense.  

However, I actually do not have any idea about how to use  `ini(parameter=after())`.  

I also think it should be the same between model piping and ini piping and I don' t think this interface does this.

mattfidler:
This probably needs to be revisited.

Note there is a possibility to also use the models to overwrite labels too.  It probably should be an option.

Right now it will be turned off:

https://github.com/nlmixr2/rxode2/blob/96c692776aba3bbf779e954c84218fe2486afd4c/R/piping.R#L110-L115

billdenney:
Should I wait until the new `ini(model)` in #415 is merged before jumping back to this one?

mattfidler:
> Should I wait until the new `ini(model)` in #415 is merged before jumping back to this one?

No.  Not needed.  If yours works, I will simply add the option myself.

billdenney:
@mattfidler, if the tests pass, then I think that this is ready for review.  The main issues that I have with this code are:

1. It doesn't handle unquoted parameter names.  (I spent some time trying to figure that one out, and I couldn't make it work going up more than one level to find the correct quoted name.)
2. An idiosyncrasy that multiple parameters are inserted in reverse order.  This is because the `append` argument is not updated with subsequent parameters.  This would be preferable to have it keep the ordering, but I couldn't see a simple way to track the append argument with the one-line-at-a-time way that the code works now.  A significant rewrite of the `ini()` generic could work around this, but that would also hit `nlmixr2est()`, so I thought I'd pause at this level.

billdenney:
@mattfidler , any ideas why this is segfaulting?

mattfidler:
Nope.  Thats why I paused on my pull request.  Perhaps it is caching
rxode2random to some odd version.

For me, the piping of models works just fine and checks perfectly fine with
the ubuntu environment.

billdenney:
You can try resetting the cache by giving it a new version: https://github.com/r-lib/actions/tree/v2/setup-r-dependencies

mattfidler:
Yep ir was the cache.

mattfidler:
Now this crashes for me on my computer....

mattfidler:
Should this be changed?
```
- Parameter order may be modified via `ini(param = after(param2))`
```

billdenney:
Yeah, I'll fix that in a push directly to main.

PR #402, opened 2023-01-13
Crash related to the fix-190 branch

billdenney:
This is an issue report that is more of a bug (i.e. don't merge it).  I can't replicate it another way.

To replicate, please do the following:

* `devtools::load_all()`

run the following:

```
  one.cmt <- function() {
    ini({
      tka <- 0.45
      tcl <- log(c(0, 2.7, 100))
      tv <- 3.45
      eta.ka ~ 0.6
      eta.cl ~ 0.3
      eta.v ~ 0.1
      add.sd <- 0.7
    })
    model({
      ka <- exp(tka + eta.ka)
      cl <- exp(tcl + eta.cl)
      v <- exp(tv + eta.v)
      linCmt() ~ add(add.sd)
    })
  }

  d <- nlmixr2data::theo_sd
  d$SEX <- ifelse(d$ID < 7, "M", "F")
  d$fSEX <- factor(d$SEX)

  sim <- rxSolve(one.cmt, events = d, keep = c("SEX", "fSEX"))
  expect_type(sim$SEX, "character")
  expect_s3_class(sim$fSEX, "factor")
  expect_equal(levels(sim$fSEX), c("F", "M"))
```

PR #406, opened 2023-01-15
Switch to arma_cerr_stream<T>

eddelbuettel:
Armadillo updates its coding convention over time. We have been fairly hard at work to have RcppArmadillo-using packages switch from a now-deprecated older way of instantiatig variables to a newer one (see [issue #391](https://github.com/RcppCore/RcppArmadillo/issues/391) for details; this PR is part of the related [issue #402](https://github.com/RcppCore/RcppArmadillo/issues/402)).

When compiling packages using RcppArmadillo with the deprecation-warning-suppressor we still use, some new warnings come up. One concerns a deprecated way of setting an output or error stream as your package does in one spot in one file. Changing this is fairly straightforward, and affects only that one file.

The package compiles with the change as it did before; I could not test it as it didn't want to load in the quick check I made.  I trust that this is minor as the PR change is.

It would be much appreciated if you could apply this pull request and update the package at CRAN within the next few months. Let me know if you have any questions.

### DIFF
--- a/src/expm.cpp
+++ b/src/expm.cpp
@@ -326,7 +326,7 @@ extern "C" int indLin(int cSub, rx_solving_options *op, double tp, double *yp_, 
   // double phiTol=op->indLinPhiTol;
   // double phiAnorm = op->indLinPhiAnorm;
   std::ostream nullstream(0);
-  arma::set_cerr_stream(nullstream);
+  arma::arma_cerr_stream<char>(&nullstream);
   
   int locf=(op->is_locf!=2);
   double tcov = tf;


mattfidler:
Thanks @eddelbuettel

eddelbuettel:
You're extra speedy as usual -- appreciate it!

conradsnicta:
@mattfidler Suggest not to submit to CRAN just yet. There is a problem with the the use of `arma_cerr_stream<T>`. Details at: https://github.com/RcppCore/RcppArmadillo/issues/402#issuecomment-1384670523

mattfidler:
Ok @conradsnicta ,

Let me know when it is ok.

conradsnicta:
@mattfidler Removing all use of `arma::set_cerr_stream()` should be sufficient. No replacement code should be necessary.

Armadillo by default prints far fewer warnings since version 10.4, so use of `arma::set_cerr_stream()` is not required.

More details at: https://github.com/RcppCore/RcppArmadillo/issues/402#issuecomment-1384816643

mattfidler:
Ok.

mattfidler:
Thanks again

eddelbuettel:
It may make sense to cook up an example (in a gist or blog post or contributed example) where we show how to suppress a message in the new way.  If someone has an example ....

mattfidler:
True, I think it was `ARMA_WARN_LEVEL` pre-declaration, that is `#define ARMA_WARN_LEVEL 0`

Is that right @conradsnicta

eddelbuettel:
It would also be nice (but maybe not easily possible) to have a _run-time_ handle and/or a more local handle than a global, whole-project compiler flag.

PR #425, opened 2023-02-01
Simplify and add testing for .amtTrans()

billdenney:


mattfidler:
Fix the CodeFactor error and we will be good to go here.

Note that I have been asked by CRAN to fix these `&` issues before (just FYI)

PR #436, opened 2023-02-07
Move getValue to header in preparation of lincmt

mattfidler:


mattfidler:
Check that bioavailability is only defined once per compartment

mattfidler:
Note that we need will need the derivatives at the time when the infusion turns on and off, so the linear compartment model should really also include when the infusion turns off in its internal data too.

mattfidler:
Abandoned.

PR #462, opened 2023-03-07
459 it is possible to set an estimate outside of the parameter boundaries

mattfidler:


billdenney:
That was just how I was thinking of it.  Looks good to me.

PR #470, opened 2023-03-11
Add a body assignment method for rxUi objects

billdenney:
This replaces nlmixr2/nlmixr2lib#33 with a slightly more general implementation that can be shared across many parts that may be able to use it.

For example, by putting it here, we could use it in `nlmixr2lib` for its function modification and `nlmixr2extra` to simplify some of the covariate search code.


R/function-body.R
```diff
@@ -0,0 +1,109 @@
+setOldClass("rxUi")
+
+bodySetRxUi <- function(fun, envir = environment(fun), value) {
+  if (is.function(value)) {
+    value <- body(value)
+  }
+  dropEnv <-
+    c(
```
mattfidler, line 8: These are hard coded.... I don't think they should be, it should be a white list not a black list

mattfidler: Or should be allowed to be removed by some heuristic.

mattfidler (commented):
This simply "modifies" the body and allows things to be retained inside of it?

R/function-body.R
billdenney: @mattfidler, I don't know what the correct list should be.  I copied this from nlmixr2/nlmixr2lib#33.  If you have a better way to do it, can you please let me know what that is specifically (or make a commit to this PR to do it)?

billdenney:
> This simply "modifies" the body and allows things to be retained inside of it?

Yes.  The simple way to think of it is that you can change the model itself (`ini()` and `model()`) while retaining everything around it like data, estimation methods, etc.

From my perspective, this is how I'd like to handle any function modification methods for nlmixr2/rxode2:

1. pull the function out with `as.function()`
2. modify the extracted function by traversing the parse tree and making any desired modifications there
3. replace the function body back with `body(ui) <- newFunction`

mattfidler:
Its failing the environment is defunct on my version of ubuntu

billdenney:
I was trying to isolate that.  I didn't think of interactive vs batch mode.

mattfidler:
I think it is because `environment(fun)` can return `NULL`

mattfidler:
However according to docs:

```
     If ‘fun’ is a function or a formula then ‘environment(fun)’
     returns the environment associated with that function or formula.
     If ‘fun’ is ‘NULL’ then the current evaluation environment is
     returned.
```

So it shouldn't...

mattfidler:
Failed on my system.

mattfidler:
This will not work in the long run; 

see:

https://github.com/wch/r-source/blob/e939a5b8c0c942e23efb9c26a1f7229d929504c8/src/library/base/R/formals.R#L33-L42

It will be taken away.  There is also no way to suppress the warnings that are happening.

mattfidler:
The problem is in the `body<-` in the R code I believe

billdenney:
Thank you for taking this forward.  I was at a bit of a loss, and it seems like your impression was similar to mine about how this was turning to be near-impossible using `body<-`.

mattfidler:
This is ready Bill; now I need to rewrite most of `nonmem2rx` handling.

mattfidler:
Nevermind: not ready; looking at `nonmem2rx` I also need rename handling.

mattfidler:
It is mostly done now, I think.

The API is important and this may be a bit easier to understand for many people, so it may be an overall improvement for most.  Though the timelines make it tight.

Before it is done, I guess I should check for proper input and allow strings to be input as well as quoted expressions.

Also maybe `rxUiCompress()`/`rxUiDecompress()` should be more careful of the object class.

The `$sticky`  needs to be documented somewhere, as with everything else... But for now it will be a hidden `API` item.

mattfidler:
Need to merge to test the other packages.

PR #482, opened 2023-03-20
zeroRe() works with correlated omega matrices

### LINKED ISSUE #480: `zeroRe()` fails with correlated etas
I will fix this.

``` r
library(rxode2)
#> rxode2 2.0.12 using 8 threads (see ?getRxThreads)
library(nlmixr2lib)

mod <-
  readModelDb("PK_1cmt") %>%
  addEta(c("lcl", "lvc")) %>%
  ini(etalcl + etalvc ~ c(0.1, 0.05, 0.2))
#> ℹ parameter labels from comments are typically ignored in non-interactive mode
#> ℹ Need to run with the source intact to parse comments
#> ℹ parameter labels from comments are typically ignored in non-interactive mode
#> ℹ Need to run with the source intact to parse comments
#> ℹ promote `etalcl` to between subject variability with initial estimate 0.1
#> ℹ change initial estimate of `etalcl` to `0.1`
#> ℹ promote `etalvc` to between subject variability with initial estimate 0.1
#> ℹ change initial estimate of `etalvc` to `0.1`
#> ℹ parameter labels from comments are typically ignored in non-interactive mode
#> ℹ Need to run with the source intact to parse comments
#> ℹ change initial estimate of `etalcl` to `0.1`
#> ℹ add covariance between `etalvc` and `etalcl` with initial estimate `0.05`
#> ℹ change initial estimate of `etalvc` to `0.2`
zeroRe(mod)
#> Error: vectors and list need to named numeric expression
```

<sup>Created on 2023-03-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

billdenney:
Fix #480

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,8 @@
+# rxode2 development version
+
+- A bug was fixed so that the `zeroRe()` function works with correlated omega
+  values.
+
 # rxode2 2.0.12
 
 ## New features
--- a/R/piping-ini.R
+++ b/R/piping-ini.R
@@ -625,15 +625,23 @@ zeroRe <- function(object, which = c("omega", "sigma"), fix = TRUE) {
   # In the code below there is no test for bounds since the bounds are typically (0, Inf).
   if ("omega" %in% which) {
     if (fix) {
-      fmtOmega <- "%s ~ fix(0)"
+      fmtOmega <- "%s ~ fix(%s)"
     } else {
-      fmtOmega <- "%s ~ 0"
+      fmtOmega <- "%s ~ c(%s)"
     }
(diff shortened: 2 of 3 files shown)


mattfidler:
Just curious, why not simply modify `iniDf` directly?  It would likely be faster for the joint case since you don't have to compress/decompress more than once.

billdenney:
I think that I had `ini()` on the brain.  Modifying `iniDf` would be simpler.

mattfidler:
I will let you think about if you want to change it.

As long as it works it doesn't much matter to me.

When you are used to using a hammer, you see everything as a nail, I guess.

PR #574, opened 2023-08-30
Allow rxAppendModel() when the models share no variables

### LINKED ISSUE #573: Allow `rxAppendModel()` when no variables are shared
Related to #520

I wanted to combine two PK models so that they could be used together in a single PD model.  By design, the two PK models share no common variables or states.  When I tried to combine them, it failed with the following message:

```
> rxode2::rxAppendModel(prep_pk_1, prep_pk_2)
# Error: the first model does not have variables that are used by the second model
```

I think that it is best to give that error by default, but I think that it would be useful to be able to override it.

billdenney:
fix #573

This is untested because I don't have the full development stack of nlmixr2-verse packages right now.  Also, documentation will need to be rebuilt.

### DIFF
--- a/R/ui-bind.R
+++ b/R/ui-bind.R
@@ -1,21 +1,21 @@
-#' Append two rxui models together
+#' Append two rxUi models together
 #'
-#' @param model1 rxUi model 1
-#' @param model2 rxUi model 2
+#' @param model1,model2 Two rxUi models to combine
+#' @param requireShared Require at least one shared variable between model 1 and
+#'   2
 #' @return New model with both models appended together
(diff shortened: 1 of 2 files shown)


mattfidler:
This is superseded by #618 

There is more done here than simply allowing common models.  It is a bit more robust than the current implementation allowing missing `ini` blocks as well as duplicate ini parameters (in most circumstances except when that parameter is involved in an `omega` covariance)

mattfidler:
I will let you delete it so you can resurrect it if you think something needs to be added.

PR #585, opened 2023-09-18
584 rbind for rxsolve

mattfidler:


mattfidler:
Probably need some provision for solving when the structure events etc are exactly the same.

PR #635, opened 2023-12-12
Move Rinternals inclusion order down

mattfidler:
Probably need to get gcc san working to double check


mattfidler:
Gcc san worked correctly waiting for CRAN.

mattfidler:
Added to CRAN

PR #662, opened 2024-03-07
661 make sure the UI is uncompressed before checking for meta information

mattfidler:


mattfidler:
Likely need to fix the cache.

Issue #730, opened 2024-07-19
covariates with different extrapolation method


motyocska:
Hi Matt,

thought I would ask: is it possible to specify covariates with different extrapolation method? Ie have weight for example as locf and creatinine as linear within the same model? It sems that this is possible if I do the extrapolation outside of the event table then add et$cov to table that is already extrapolated with the specific method but what I would like is to have rxSolve do the extrapolation work instead... 

thanks,

Andras

billdenney:
My method for linear interpolation is to do it manually by adding 3 columns to the dataset (4 columns total):

* Current covariate value
* Current time (already there as your "time" column)
* Next covariate value
* Next covariate time

Then, I manually interpolate.

My guess is that you may already do this, but hopefully it helps someone else.

motyocska:
thank you Guys!

mattfidler:
Since I have been changing covariate handling with `iCov`, this is probably a good time to change this; 

From the release notes:

- Now you can specify the interpolation method per covariate in the model:

  - `linear(var1, var2)` says both `var1` and `var2` would use linear
    interpolation when they are a time-varying covariate. You could
    also use `linear(var1)`

  - `locf()` declares variables using last observation carried forward

  - `nocb()` declares variables using next observation carried backward

  - `midpoint()` declares variables using midpoint interpolation


Once #749 is merged you could use the develpment branch to specify different interpolation methods with whatever model you want.

motyocska:
thanks Matt! Loking forward to try it!
Andras Farkas, Pharm.D 
    On Saturday, August 3, 2024 at 04:05:43 PM EDT, Matthew Fidler ***@***.***> wrote:  
 
 


Closed #730 as completed via #749.

mattfidler:
Great @motyocska

PR #752, opened 2024-08-04
Have simulationModel and family respect locf() types of rules

mattfidler:
Related to new feature #730


mattfidler:
Perhaps also need to add piping directives in `model({})`

For example:

- ` |> model(-locf)`
- ` |> model(locf(vars))`

PR #782, opened 2024-10-10
Add user function in rxode2 ui that can change code when evaluating

mattfidler:
Fore example

`linMod(x, 3)`  

in the code base would translate to:

`rxLM1_1 + rxNM1_2*x + rxNM1_3*x^2 +  + rxNM1_4*x^4`

And all the new constants would be in the `ini` file after evaluating the function.


mattfidler:
For this to be more useful it also will require:
- [ ] variable number arguments
- [ ] named arguments
- [ ] `lhs` of the current expression

And in conjunction with `nlmixr2est` would require:

- [ ] A user function that can see the data to be fit/simulated.

PR #799, opened 2024-10-24
Documentation and pkgdown fix

billdenney:
No functional changes


billdenney:
@mattfidler, I got things farther, but there seems to be something like a crash here. Can you diagnose that?

mattfidler:
It isn't a crash per-se it a difference in how `evaluate` behaves with the new version.  I simply need to completely rewrite `et()` or take out the examples.  It consistently hangs when generating these examples though they run interactively just fine.

mattfidler:
I think this is an error in evaluation in either `pkgdown` or `evaluate`.  Ignoring these works just fine.

PR #800, opened 2024-10-24
simplify repeated ifelse

### LINKED ISSUE #798: Simplify repeated `ifelse`
https://github.com/nlmixr2/rxode2/blob/a19a30c925ff1726af9d184c7397a1916983cc10/R/et.R#L1325-L1345

https://github.com/nlmixr2/rxode2/blob/a19a30c925ff1726af9d184c7397a1916983cc10/R/et.R#L1349-L1369

billdenney:
Fix #798


mattfidler:
I agree the code is more readable.

However, this makes `dplyr` an `Imports` instead of a `Suggests`; we are/were close to the 20 threshold so I don't know if this is a good idea because of CRAN requirements.

Issue #689, opened 2024-06-12
`digest` method for most/all object types


billdenney:
I'm working on `nlmixr2targets`, and I find that in some scenarios, I consistently get different model digest values (`digest::digest(rxuiObject)` differs between R sessions for rxUi objects that are created the same way).  I think that this relates to creating `digest`s of the environments or temporary files.

That said, the `rxuiObject$md5` value remains consistent between R sessions.

1. Would it be reasonable to create a `digest.rxUi` method that would return the `object$md5` value or can you think of a reason that should not be done?
2. And related to that, would it make sense to create a `digest.nlmixr2FitCore` method that would do the same thing?
3. Would there be any other object types that should have a `digest` method created?
4. Should this live in the specific packages (`rxode2::digest.rxUi`, `nlmixr2est::digest.nlmixr2FitCore`) or within `nlmixr2targets`?

billdenney:
This is related to #163.

billdenney:
@mattfidler, I was trying to trace where `digest` values differ when I'm creating what I think is the same object. I found in an `rxUi` object that the `ui$mv0$timeId` value differs. It appears to be some type of timestamp. Is it reasonable to ignore that when creating a digest of an `rxUi` object or is it important in some way for the ui object?

billdenney:
I think that the following would provide a stable `digest` method by ignoring compilation time:

```
sha1.rxUi <- function(object, ...) {
  rxuitmp <- object
  rxuitmp$mv0$timeId <- -1L
  digest::sha1(unclass(rxuitmp))
}
```

But, it may break things if assignment of `-1L` changes the original model.

billdenney:
This is outside of `rxode2` but related. For an `nlmixr2FitData` object one element that differs as well is the `object$env$time` data.frame which gives duration of the fit steps. That should probably be omitted from its `sha1` method.

billdenney:
From https://github.com/nlmixr2/rxode2/pull/835#issuecomment-2569380983
> I think that possibly part of the issue is the definition of "same". Rather what should give the same hash and what shouldn't.

I agree that the definition of "same" is an issue here.

I'm going to try to be more explicit about what I'm looking to solve.

I'm trying to make `nlmixr2targets` accurately identify when a model needs to be run again or not. `targets` uses the `digest()` function to determine if an R object has changed. If an object that is an input to a target has changed (as determined by `digest()` from a prior to current `tar_make()` run, the target will rerun.

[...]

PR #835, opened 2025-01-01
Create `sha1()` for stable `digest()` results

billdenney:
I'm having trouble identifying which parts of the rxUi object should be stable between calls and which are likely to change. What I have here is definitely *not* the right solution, but it is code worth discussing.


billdenney:
@mattfidler, The challenge that I have is that when I try to create models that should be the same, the results are not always the same. I have tried to track down which parts differ, and I get different answers each time. Can you give guidance about what parts are required for testing equivalence when using a model again vs which parts could be ignored?

R/digest.R
```diff
@@ -0,0 +1,39 @@
...
+  for (nm in nmCheck) {
+    objectSha1[[nm]] <- sha1(object[[nm]], digits = digits, zapsmall = zapsmall, ..., dropSessionSpecific = dropSessionSpecific, algo = algo)
+  }
+  sha1(objectSha1, digits = digits, zapsmall = zapsmall, ..., dropSessionSpecific = dropSessionSpecific, algo = algo)
+}
+
+#' @export
+sha1.rxModelVars <- function(x, digits = 14L, zapsmall = 7L, ..., dropSessionSpecific = TRUE, algo = "sha1") {
+  nmCheck <- names(x)
+  if (dropSessionSpecific) {
```
billdenney, line 20: @mattfidler, These object parts are some of the places where I have sometimes seen differences between the same model, and sometimes I don't.

R/rxode-options.R
```diff
@@ -187,26 +187,42 @@ rxTempDir <- function() {
...
   .tmp <- .normalizePath(.tmp)
   Sys.setenv(rxTempDir = .tmp)
   utils::assignInMyNamespace(".rxTempDir0", .tmp)
   utils::assignInMyNamespace("rxode2.cache.directory", .tmp)
   invisible()
 }
+#' @describeIn rxCreateCache Remove the cache directory for rxode2 and unset
+#'   related variables
+#' @export
+rxRemoveCache <- function(subdir = "") {
```
billdenney, line 216: @mattfidler, I created this function to try to force a model to be compiled differently. But, it didn't always work.

tests/testthat/test-digest.R
```diff
@@ -0,0 +1,63 @@
...
+      d/dt(peri)  <- Q*C2 - Q*C3
+      eff(0) <- 1
+      d/dt(eff)   <- Kin - Kout*(1-C2/(EC50+C2))*eff
+    })
+  }
+
+  ui1 <- rxode2(mod1)
+  ui1$mv0$timeId
+
+  # Wait a few seconds so that compilation will occur at a different time
```
billdenney, line 27: @mattfidler, Is there any way to force an `rxUi` object to recompile, even if it is otherwise identical to an existing model? Recompilation seems like a part of making an `rxUi` object different.

tests/testthat/test-digest.R
```diff
@@ -0,0 +1,63 @@
...
+      d/dt(depot) <- -KA*depot
+      d/dt(centr) <- KA*depot - CL*C2 - Q*C2 + Q*C3
+      d/dt(peri)  <- Q*C2 - Q*C3
+      eff(0) <- 1
+      d/dt(eff)   <- Kin - Kout*(1-C2/(EC50+C2))*eff
+    })
+  }
+
+  ui2 <- rxode2(mod2)
+  ui2$mv0$timeId
```
billdenney, line 53: @mattfidler, I added `ui1$mv0$timeId` and `ui2$mv0$timeId` to try to make the models different. It didn't always work.

tests/testthat/test-digest.R
```diff
@@ -0,0 +1,63 @@
...
+
+  ui1 <- rxode2(mod1)
+  ui1$mv0$timeId
+
+  # Wait a few seconds so that compilation will occur at a different time
+  Sys.sleep(2)
+
+  mod2 <- function() {
+    ini({
+      KA=0.294
```
billdenney, line 32: @mattfidler, I used normal numbers rather than SI notation to try to make the models compile differently, but it didn't work.

tests/testthat/test-rxode-options.R
```diff
@@ -0,0 +1,34 @@
+test_that("rxCreateCache", {
```
billdenney, line 1: @mattfidler, If caches are used during model testing or otherwise, this could cause models to recompile. These tests may be better to skip most of the time.

R/digest.R
mattfidler: Hi @billdenney 

This is contained in the model variables already. If it isn't working there, we need to refactor it.  It has implications on compiling and cashing.  I think this method shouldn't not be added (as written).

R/rxode-options.R
mattfidler: I'm unsure what you are trying to accomplish here, and why, perhaps some context would be helpful

tests/testthat/test-digest.R
mattfidler: You can but not with an exported method, and not very well.

tests/testthat/test-digest.R
mattfidler: This shouldn't be where the sha should be queried, if it needs to be done.

tests/testthat/test-digest.R
mattfidler: This is because R handles the numbers and SI vs normal numbers are stored in R exactly the same, so it shouldn't have differences.  This is expected.

Also rxode2 uses normalized model syntax to compare models.  So

```
x^2
```

Is equivalent to

```
x**2
```

Which it should be since it generates the same code :smile:

tests/testthat/test-rxode-options.R
mattfidler: While I am all for testing, this would definitely break parallelism (if ever implemented)

mattfidler:
One of the bigger comments I have about this, is that rxUi models are not deterministic any longer. 

With the addition of the rxui model functions (to support NN ODEs), the user can provide values that will not always give the same model.  In particular the NN ODEs use random initial value of weights satisfying a regulatory condition.  Hence running the same model may give different results every time.

I have thought about trying to use `withr` to save the seed (though I think this actually puts us over 20 Imports) to make all models deterministic.  This is what is done with SAEM (which also suffers from the lack of a deterministic solution, but it sets the seed directly instead).  I could do that as well, though I prefer to keep away from setting the users seed.

The things that determine the hash of the model should be:

- The hash of the model variables
- The lines of the model representing the residual errors
- The initial values included in the model

billdenney:
This PR isn't the right solution.

I'm looking for something that will give a deterministic `digest()` for use with `nlmixr2targets`. That deterministic digest should encompass all user specified model parts, but non-deterministic parts (like NN initial weights) would ideally not be considered.

This is a problem that I have tried to solve several times without success.

mattfidler:
I think that possibly part of the issue is the definition of "same".

billdenney:
I'm going to close this PR and move the conversation back to issue #689. This PR is not the solution to the problem. (The changes to the cache functions were to try to find a way to make models be the same or differ. They aren't necessary otherwise.)

PR #856, opened 2025-04-15
827 linear compartment solutions

mattfidler:


mattfidler:
The `linCmtB` solutions are slower than odes.

mattfidler:
- Creating memory
- transposing
- macros2micros

mattfidler:
There are only 'skip'ped commits left to test.
The first bad commit could be any of:
88bdbf79be3895d3aab1777702ac98d45364cc5b
b0ed2d1da71df697318a9b022beba86badcc3d9f
be5d450d41d47ab8bd576e8c86aa033bf93c753b
We cannot bisect more!

mattfidler:
There are only 'skip'ped commits left to test.
The first bad commit could be any of:
69592635e5b31f3f89344a60c387104a51d96981
c25a1cd6e5750903674ded2847197e7d3c89b0dd
We cannot bisect more!

mattfidler:
Now with `linCmtB` the number of sensitivities requested will be honored, meaning that if you are using a 2 compartment model with focei and only have a between subject variability on clearance, it will be faster than having between subject variabilities on every parameter.

``` r
devtools::load_all("~/src/rxode2")
#> ℹ Loading rxode2
#> rxode2 4.0.0 using 11 threads (see ?getRxThreads)

f <- function(CL=7.5, V1=20, V2=30, V3=200, Q2=0.5, Q3=0.05, ka=1, DOSE=100, tau=24, deriv=FALSE, ndiff=127) {
  p1 <- CL
  v1 <- V1
  p2 <- Q2
  p3 <- V2
  p4 <- Q3
  p5 <- V3
  ka <- ka
  if (deriv) {
    alastNV <- rep(0, 26)
  } else {
    alastNV <- c(0, 0, 0, 0)
  }
  rateNV <- c(0, 0)
  oral0 <- 1L
  trans <- 1L
  ncmt <- 3
  type <- 3L # bolus infusion
  amt <- DOSE
  dt <- 0
  cmt <- 0L
  tinf <- 0
  .Call(`_rxode2_linCmtModelDouble`, 0.1,
        p1, v1, p2, p3, p4, p5, ka,
        alastNV, rateNV, ncmt, oral0, trans,
        deriv, type, tau, tinf, amt, cmt, ndiff)
}

f.ndiff <- function(n) {
  if (n==7) return(127)
  sum(2^sample(0:6, size=n))
}

fn <- function(n){
  f(deriv=TRUE, ndiff=f.ndiff(n))
}




library(microbenchmark)
library(ggplot2)

mb <- microbenchmark(fn(1),
                     fn(2),
                     fn(3),
                     fn(4),
                     fn(5),
                     fn(6),
                     ## fn(7), # this is biased by the sampling
                     times=100)


autoplot(mb) + xgxr::xgx_scale_y_log10() +
  rxTheme()
#> Scale for y is already present.
#> Adding another scale for y, which will replace the existing scale.
```



<sup>Created on 2025-05-08 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

PR #1017, opened 2026-04-20
Serialise assignFuns() in ind_solve when called from a parallel region

billdenney:
ind_solve() calls assignFuns(), which on first call routes through _assignFuns0() → many R_GetCCallable lookups (R's symbol table, not thread-safe). nlmixr2est's FOCEi inner loop calls ind_solve from its own parallel-for, so multiple threads race on the first-call path. Gates the call with omp critical when omp_in_parallel() is true; zero overhead otherwise. A stronger fix (critical section inside the generated _assignFuns(), or CAS on _assign_ptr) is tracked as a follow-up.


mattfidler:
This is split out because it isn't being called in parallel...  Having the parallel guard here is unnecessary.  Perhaps it is focei.  I will check.

mattfidler:
This is used in inner. Fixed issue flagged by check.

mattfidler:
This doesn't work in general, but it is a good start.

mattfidler:
I am going to close this; it protects and assigns just once.  To allow the right assignment, the structure needed to be changed as in #1036

PR #1026, opened 2026-04-29
Allow split bolus doses

mattfidler:


mattfidler:
@billdenney this could allow:

- A model library of complex absorption types without modifying the data
- For instance you can use `splitBolus(depot1, depot1, depot2)` in the model to dose to both compartments.

I can't see a reason to split infusions or any other types of events.

PR #1031, opened 2026-05-01
Range-check dparse input length to prevent (int)strlen truncation

billdenney:
`tran.c:572` feeds the model source to dparser via
    _pn = dparse(curP, gBuf, (int)strlen(gBuf));
When `gBuf` exceeds INT_MAX bytes the `(int)` cast silently wraps to a wrong (often negative) length and dparser reads past the buffer.

Wrap the call so the `strlen` result is computed in `size_t`, checked against `(size_t)INT_MAX`, and only then cast to `int`.  Out-of-range input gets a clean R error instead.

Adds tests/testthat/test-mem-dparse-int-cast.R as a regression test. The boundary test is `skip()`ed because it requires ~3 GB of free RAM.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,10 +1,10 @@
 # rxode2 (development)
 
-- Switch `tran.c` parser entry from `dparse(curP, gBuf, (int)strlen(gBuf))`
-  to `udparse(curP, gBuf, (unsigned int)strlen(gBuf))`.  The new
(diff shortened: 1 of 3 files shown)


mattfidler:
Again, this requires ABI changes in dparser and I don't really feel like bringing down nlmixr2 right now

PR #1038, opened 2026-05-04
Per-individual neqOverride C-API + plumb through getSolve/getAdvan

billdenney:
## Summary

- Adds `setIndNeqOverride()` / `getIndNeqOverride()` C-API at pointer
  table slots 56/57 (and as `R_RegisterCCallable` entries) plus a
  `neqOverride` field on `rx_solving_options_ind` (default `-1`).
- Adds `rxEffNeq(ind, op)` static-inline helper in `inst/include/rxode2.h`
  that returns `ind->neqOverride` when set in `[0, op->neq]`, else `op->neq`.
- Routes `getSolve()`, `getAdvan()`, and `getOpIndSolve()` through the
  helper so a per-individual pred-mode solve writes/reads at the compact
  stride without mutating the shared `op->neq` from a parallel worker
  thread.

## Why

Closes the half-fixed gap behind nlmixr2est PR #621 (parallel FOCEi):
`shi21EtaGeneral` mutates `op->neq` to switch from inner-mode to
pred-mode, which races with concurrent inner solves on other threads
and produces heap corruption in `lsoda_free` for models whose `f()`,
`dur()`, `rate()`, or `lag()` depend on ETAs (Matt Fidler's 2026-05-01
review comment on PR #621).

This PR exposes the metadata; nlmixr2est consumes it via its companion
PR (parallel-v2). Allocations remain sized for `op->neq`; the override
is constrained to `<= op->neq` by caller contract.

## Test plan

- [ ] Existing rxode2 tests (`devtools::test()`) pass — additive change.
- [ ] nlmixr2est `tests/testthat/test-focei-parallel.R` passes with
      cores=2 baseline / mceta=0 / mceta=2.
- [ ] `inst/reprex_fbio_eta_parallel.R` in nlmixr2est (the f(depot)+ETA
      reprex) survives cores=2 once nlmixr2est PR #621 lands on top.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,12 @@
 # rxode2 (development)
 
+- Add `getIndNeqOverride()` / `setIndNeqOverride()` C-API at pointer
+  table slots 56 and 57 (and as `R_RegisterCCallable` entries).  This
+  allows downstream packages (e.g. nlmixr2est) to mark a per-individual
+  effective neq for parallel-FOCEi pred/inner alternation without
+  mutating the shared `op->neq` from a parallel worker thread.  Default
+  value is -1 (use `op->neq`).
+
 - Add `evid_()` function to allow arbitrary doses and observations in
   a rxode2 model.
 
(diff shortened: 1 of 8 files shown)


mattfidler:
While I know why this would be needed, it isn't explained very well in the log.  

In fact the AI news items will be stripped eventually because the level of detail is too much for the audience (I think).

billdenney:
I updated the historical record with the AI-suggested info in the title and initial description.

PR #1042, opened 2026-05-07
Num compare

mattfidler:


mattfidler:
This doesn't do what was requested (at all)

PR #1056, opened 2026-05-27
Declare censoring aesthetics in stat_cens, tighten pair validation

### LINKED ISSUE #31: `geom_cens()` uses aesthetics of `CENS` and `LIMIT` rather than `lower` and `upper`
This line:

https://github.com/nlmixr2/nlmixr2plot/blob/85b40df339f33196a7364e6f71903cf8a900c561/R/plot.nlmixr2.R#L284

Causes warnings of:

```
1: In geom_cens(ggplot2::aes(lower = .data$lowerLim, upper = .data$upperLim),  :
  Ignoring unknown aesthetics: lower and upper
```

billdenney:
Add `optional_aes = c("cens", "limit", "lower", "upper")` to `StatCens` so ggplot2 stops emitting "Ignoring unknown aesthetics" warnings when callers map the censoring aesthetics. The internal `.setupGroupCensBox()` already reads these column names; they were just never declared on the ggproto.

Tighten pair validation in `.setupGroupCensBox()`:

  - reject mixing the (cens, limit) and (lower, upper) forms (previously only the cens+lower combination errored, and with a misleading "lower aesthetic" message even for the upper check)
  - require both `lower` and `upper` together (previously `upper` alone was silently accepted and produced empty output)
  - reject `limit` without `cens` (previously silently ignored)
  - require one of the two pair forms (previously empty data passed through)

`cens` alone (with implicit -Inf/Inf bounds) remains accepted.

Fix the roxygen docs to describe the actual lowercase aesthetic forms. The previous text claimed `CENS`/`LIMIT` (uppercase), which the implementation has never accepted.

Fixes nlmixr2/nlmixr2plot#31

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: rxode2
-Version: 5.1.1
+Version: 5.1.2
 Title: Facilities for Simulating from ODE-Based Models
 Authors@R: c(
    person("Matthew L.","Fidler", role = c("aut", "cre"), email = "matthew.fidler@gmail.com", comment=c(ORCID="0000-0001-8538-6691")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,13 @@
 # rxode2 5.1.1
 
+- `geom_cens()` / `stat_cens()` no longer emit "Ignoring unknown
+  aesthetics" warnings when censoring aesthetics are mapped.
+  Documentation corrected to describe the two supported lowercase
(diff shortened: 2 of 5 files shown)


billdenney:
@mattfidler, Any issue with merging this?

mattfidler:
Mostly I would prefer to wait until rxode2 hits CRAN.  This will likely not be part of that release.

billdenney:
It's not a rush to get this into the next CRAN release.

PR #1064, opened 2026-06-06
Make atolRtolFactor_ and the parallel liblsoda solve thread-safe

billdenney:
## Summary

`atolRtolFactor_` mutated the shared `_globals.gatol2 / grtol2` arrays
and `op->ATOL / op->RTOL` while parallel OpenMP workers were reading
them through `op->atol2 / op->rtol2`. When two workers hit bad solves
concurrently they raced on the writes, leaking timing into downstream
optimizer trajectories (most visibly in `nlmixr2est` FOCEI, where same
seed → different fitted parameters across runs).

This PR splits the tolerance arrays into per-thread slices indexed by
`omp_get_thread_num()`:

- New `_globals.gatol2Thread / grtol2Thread` sized `cores * neq`, sliced
  out of the existing `gsolve` allocation (no new malloc), with the
  total allocation size and memory-availability guards updated to match.
- Each per-thread slice is initialized from `_globals.gatol2 / grtol2`
  at solve setup, so a worker's first read sees the same base
  tolerances the previous implementation provided.
- `_setIndPointersByThread` wires `ind->atol2 / rtol2` to the calling
  thread's slice.
- `atolRtolFactor_` writes only to the calling thread's slice.
  Outside parallel regions `omp_get_thread_num()` returns 0, so
  single-threaded behavior is bit-identical to the prior version
  (`op->atol2 / rtol2` are pointed at thread 0's slice for the same
  reason — sequential paths like `par_indLin`, `par_lsoda`, `par_dop`
  keep working unchanged).
- `ind_liblsoda0` (the parallel solve helper in `par_solve.cpp`) reads
  from `ind->atol2 / rtol2` instead of the shared `op->atol2 / rtol2`.

## Test plan
- [x] `tests/testthat/test-par-solve.R` extended with a smoke test that
      exercises `atolRtolFactor_` and asserts `cores=1` vs `cores=2`
      results agree.
- [x] Existing par-solve tests pass.

## Related
Closes one of two non-determinism sources for
https://github.com/nlmixr2/nlmixr2est/issues/641.
The companion fix (`sortIds` iota ordering) is in
nlmixr2est PR https://github.com/nlmixr2/nlmixr2est/pull/new/fix-sortids-determinism.
Both are needed together; this PR alone does not eliminate the FOCEI
non-determinism in `nlmixr2est`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,15 @@
 # rxode2 (development)
 
+- Make `atolRtolFactor_` and the parallel `liblsoda` solve path thread-safe by
+  splitting the absolute/relative tolerance arrays into per-thread slices
+  (`_globals.gatol2Thread`, `_globals.grtol2Thread`, sized `cores * neq`).
+  Previously `atolRtolFactor_` mutated the shared `_globals.gatol2/grtol2`
+  arrays while parallel worker threads were reading from `op->atol2/op->rtol2`,
+  causing non-deterministic optimizer trajectories in downstream packages like
+  `nlmixr2est` (FOCEI). The per-thread slices are written by
+  `omp_get_thread_num()`, so single-threaded behavior is bit-identical to the
+  prior implementation.
+
 - Fix potential security and memory-management issues that could lead to crashes or undefined behavior
 
 - Fix integer overflow in the internal `nSize` buffer-sizing calculation that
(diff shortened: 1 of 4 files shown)


billdenney:
This is not likely necessary

PR #1067, opened 2026-06-06
Make atolRtolFactor_ thread-safe; replace body with atolRtolFactorC_'s

billdenney:
## Summary

`atolRtolFactor_` was kept around as "NOT thread safe… nlmixr2est 5.0 needs this interface."  In addition to the race, its writes to `_globals.gatol2 / grtol2` had no effect on the parallel solve (solvers read the per-thread slices set by `_setIndPointersByThread`), so loosening calls from downstream packages were essentially silent no-ops on the actual numerics, with racy writes to `op->ATOL / op->RTOL` thrown in.

This PR makes `atolRtolFactor_` actually work, thread-safely, by replacing its body with the existing `atolRtolFactorC_` implementation:

- Per-thread writes to `_globals.gatol2Thread / grtol2Thread / gssAtolThread / gssRtolThread`
- Per-individual `ind->tolFactor` update for the subject currently solving on this thread, so `iniSubject()` reapplies the cumulative loosening regardless of which thread runs the subject on a later iteration
- No writes to `op->ATOL / op->RTOL`

`atolRtolFactorC_` is retained as a thin alias for `src/init.c`'s `R_MakeExternalPtrFn` reference.

## ABI invariance

Function signature, registration name (`_rxode2_atolRtolFactor_`), and Rcpp interface attributes are unchanged.  Existing downstream binaries that called `R_GetCCallable("rxode2", "_rxode2_atolRtolFactor_")` resolve the new code transparently when users update rxode2 — no rebuild required.  In particular, this includes the current CRAN `nlmixr2est`.

## Test plan
- [x] `tests/testthat/test-par-solve.R` source-runs to completion.
- [x] Empirically reduces FOCEI two-run OFV gap from ~1544 to ~0.77 on the https://github.com/nlmixr2/nlmixr2est/issues/641 reproducer at `cores=2`.
- [ ] Bit-identical determinism not yet achieved on that reproducer — at least one more non-determinism source remains, outside this function's scope.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,18 @@
 # rxode2 5.1.3
 
+- Make `atolRtolFactor_` thread-safe by replacing its body with the
+  per-thread / per-individual implementation that previously lived in
+  `atolRtolFactorC_`.  Loosening writes now go to the calling thread's
+  slice of `_globals.gatol2Thread / grtol2Thread / gssAtolThread /
+  gssRtolThread` and update `ind->tolFactor` for the subject currently
+  being solved on that thread, so subsequent re-solves pick up the
+  cumulative loosening regardless of which thread later runs the
+  subject.  The function signature is unchanged; existing downstream
+  binaries (including `nlmixr2est` on CRAN) pick up the fix without
+  needing a rebuild.  `atolRtolFactorC_` is retained as an alias.
+  This closes the non-determinism in parallel FOCEI optimization
+  reported in https://github.com/nlmixr2/nlmixr2est/issues/641.
+
 - Bug fix for `mix()` models as well as `iCov` models.
 
 # rxode2 5.1.2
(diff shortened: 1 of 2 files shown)


mattfidler:
This makes no sense to me.  

The Rcpp interface is out of our control, it may hand-off with a pointer lookup instead of pointer assignment on load.  The two should be equivalent.  I know that claude seems to like this piece of code being exported as Rcpp, but it never worked well in my tests.   

In theory this is a unneeded work-around.

billdenney:
There is probably not the right solution to an issue of threaded work getting a different result than single thread. The result seemed better, but it wasn't clear to me if perhaps something else was going on.

Ignore this one, since it doesn't fix the issue fully or make sense to you.

mattfidler:
Did you add a reproducible example.  The parallel solving all gave the right examples in all of the nlmixr2est tests reproducible, so I did not observe this behavior.

billdenney:
There was at least one point in time yesterday when my overall package state was messed up. This could have been transient from that rather than a real issue. The reprex was the model which needed the change to the scaling factor, but it may not have actually been an issue.

PR #1068, opened 2026-06-08
Aggregate intdy window-miss warnings instead of flooding the console

### LINKED ISSUE #301: indy printing slows model minimization
When running an oncology nonclinical model, I see messages like "intdy -- t = 504 illegal. t not in interval tcur - _rxC(hu) to tcur" many times (thousands of times-- longer than my Rstudio scroll back buffer).

When I run the model as-is, it takes 15.1 minutes.  When I run the model within `suppressMessages()`, it takes 10.3 minutes.  The specific line of code is this one:

https://github.com/nlmixr2/rxode2/blob/7342f1deb8fb3d68411b9a750870b6f5a217401e/src/intdy.c#L57

Would it be reasonable to put some form of rate-limiter on the print statement, give a flag to suppress it, or to remove the print statement to speed things up?  (I did some looking into rate limiters, but it wasn't clear that there was one best solution.)

billdenney:
Closes #301.

## Summary
- Replace the per-occurrence `RSprintf` at [src/intdy.c:56](https://github.com/nlmixr2/rxode2/blob/fix/intdy-warn-throttle/src/intdy.c#L56) with a small aggregator (`rxSolveWarnPush/Flush/Reset`, registered as C-callables and exposed in `src/solveWarn.h`). Push records `(family, subject id)` under an `omp critical` region; flush emits one summary line per family with the count and up to N distinct subject IDs (default 5, with `... (M more)` truncation).
- Carry the subject id into the LSODA context: add `int id` as the first field of `struct lsoda_common_t` so `lsoda_reset`'s `memset` (which starts after `memory`) leaves it untouched on pooled-context reuse. Set `ctx->common->id = ind->id` at the start of `ind_liblsoda0` so `intdy.c` knows which subject it's complaining about.
- Flush at the end of `par_solve` so standalone `rxSolve` users see one summary line per call.
- The literal `_rxC(hu)` that previously leaked into the user-visible message disappears with the rewrite.

## Why
Per #301, a stiff oncology model can fire the `intdy -- t = X illegal` warning thousands of times per estimation iteration, and the printing itself measurably slows the optimizer (15.1 min as-is vs 10.3 min under `suppressMessages()` in the reporter's case — ~30% wall-clock tax). The integrator default is already `liblsoda` (stiff-capable with auto-switching), so switching solvers isn't the answer — the warning is a normal symptom of a struggling integrator that just needs to be quieter.

## Companion PR
Needs `nlmixr2/nlmixr2est#new` (`fix/intdy-warn-throttle`) so the FOCEi/SAEM iteration prints invoke the new flush. The nlmixr2est side resolves `rxSolveWarnFlush` lazily via `R_GetCCallable`, so an old rxode2 still loads (with the flush no-oped) until both packages are upgraded together.

## Test plan
- [ ] `R CMD INSTALL .` clean
- [ ] `NOT_CRAN=true devtools::test(filter="solve")` no new failures
- [ ] Direct aggregator test via `R_GetCCallable("rxode2","rxSolveWarnPush"/"rxSolveWarnFlush"/"rxSolveWarnReset")`: empty flush silent, multi-id flush truncates correctly, second empty flush silent
- [ ] Standalone `rxSolve` on a stiff model collapses an intdy flood to a single line
- [ ] FOCEi fit on `theo_sd` still converges to OFV ≈ 116.99
- [ ] Reproduce the #301 oncology timing (target: closer to 10.3 min than 15.1 min)


mattfidler:
@billdenney you may want to wait for the odeint branch to be merged before trying this.

mattfidler:
It has been merged, and you may wish to make it a bit more general.

PR #1072, opened 2026-06-10
Add setRxThreadId(): let external OMP drivers supply the real thread id

billdenney:
On Windows/Rtools each package statically links its own libgomp. When an
external package (nlmixr2est's FOCEi inner loop) runs its own OpenMP team and
the workers call into rxode2's ind_solve(), rxode2's omp_get_thread_num()
returns 0 for every foreign worker, so rxode2 indexes all of its per-thread
solve buffers (LSODA ctx pool, glhs/gon/solve scratch, tolerance + dose arrays,
linCmt scratch) at slot 0 and concurrent workers write/lsoda_free the same
memory → heap corruption. Windows-only (Linux shares one libgomp.so), cores≥2.
This is the cause of the windows-latest R-CMD-check segfaults building focei fits.

Adds the C-callable setRxThreadId(int) + a thread_local override; per-thread
indexing uses it when set (≥0), else omp_get_thread_num() (correct inside
rxode2's own par_solve). Verified on Windows: ODE and linCmt focei fits at
cores=2/4/8 complete with the correct objective.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,14 @@
 # rxode2 5.1.3
 
+- Added `setRxThreadId()` so a package that drives rxode2's per-subject
+  solve from its OWN OpenMP team (e.g. nlmixr2est's FOCEi inner loop)
+  can supply the real thread id.  On Windows each package statically
+  links its own libgomp, so rxode2's `omp_get_thread_num()` returns 0
+  for every foreign worker thread, which collapsed all of rxode2's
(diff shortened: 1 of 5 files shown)


billdenney:
@mattfidler, this was a headache of a solve for some local Windows issues. I think that it covers the issues on this side.

mattfidler:
This adds ABI dependency, so this is not quite right.

PR #1073, opened 2026-06-10
Origin/claude/fix cross dll omp

mattfidler:
Not sure why I couldn't push to #1072 

This are the fixes for that branch.

@billdenney  the nlmixr2est branch needs to call the `setRxThreadId()` directly without the prefix.  In some contexts this ( `rxode2::setRxThreadId()`) has corrupted the solves in Linux at least.


mattfidler:
As I have said, claude loves this paradigm, but it doesn't work here.

PR #1074, opened 2026-06-12
Rxode2 speed 2026

mattfidler:


mattfidler:
A 2-cmt model with ini-only params and a compound coefficient like (cl+q)/v fails at model creation with a mu-referencing error ("2+ single population
  parameters in a single mu-referenced expression") — that's a pre-existing rxode2 mu-ref constraint, independent of odeToLin. Use body assignments (cl
  <- exp(tcl)) for such models.

PR #1086, opened 2026-06-20
linCmt() auto-conversion: handle ODE states coupled to the central sub-system

billdenney:
## Problem

The automatic `linCmt()` conversion (`rxSolve(..., useLinCmt = TRUE)`, the default in 5.1.3) folded every ODE state into an analytic `linCmt()` whenever a 1–3 compartment linear topology was detected. When another model line referenced a folded state as a value — e.g. a peripheral observable `Cp <- periph / vp` — the state was dropped and demoted to a required parameter, so the default solve aborted with `The following parameter(s) are required for solving: <state>`.

## Fix

`odeToLin()` now **converts when `linCmt()` can faithfully represent the system, and keeps the explicit ODE otherwise** — a UI-only transformation (no rxode2 core change):

- **Output-only coupled peripheral → convert.** The ODE compartments are renamed to their canonical linCmt names (`periph → peripheral1`, etc.) so the observable reads `linCmt()`'s solved peripheral amount, and the central endpoint is anchored to the central compartment (`Cc ~ err | central`) so no observation compartment is injected (which would otherwise make the in-model peripheral reference resolve to 0).
- **Cannot represent → keep the ODE.** A numerical mass-balance guard rejects a compartment with independent loss (a metabolite, `d/dt(met) <- kpm*central - kelm*met - kbt*met`), and a peripheral carrying its own estimated `~` endpoint is left as an ODE (it collides with linCmt's internal compartment).

## Verification

`test-odeToLin.R` 28/28. Verified against the explicit ODE solve (2-cmt IV and oral): central concentration, peripheral concentration, and depot amount all agree. Ordinary 1/2/3-cmt models still convert.

Of the four nlmixr2lib models that motivated this, the output-only-peripheral ones convert to analytic `linCmt()` while the metabolite / estimated-peripheral ones stay on the ODE solver — all solving correctly by default.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: rxode2
-Version: 5.1.3.9000
+Version: 5.1.3
 Title: Facilities for Simulating from ODE-Based Models
 Authors@R: c(
    person("Matthew L.","Fidler", role = c("aut", "cre"), email = "matthew.fidler@gmail.com", comment=c(ORCID="0000-0001-8538-6691")),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: rxode2
-Version: 5.1.3
+Version: 5.1.3.9000
 Title: Facilities for Simulating from ODE-Based Models
 Authors@R: c(
    person("Matthew L.","Fidler", role = c("aut", "cre"), email = "matthew.fidler@gmail.com", comment=c(ORCID="0000-0001-8538-6691")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,14 @@
 # rxode2 5.1.3
 
+- Fixed a UI-system bug where, in a `linCmt()` model that both has an
+  error model and reads a materialized linCmt compartment in an equation
+  (e.g. `Cp <- peripheral1 / vp`), the observation compartment injected
+  for the endpoint shifted the linCmt compartments' solved-state indices,
+  so the in-equation reference read an unwritten slot (`0`) instead of the
+  solved amount.  Such references now resolve correctly, so `central`,
+  `peripheral1`, `peripheral2` and `depot` amounts can be used in model
+  equations alongside an error model.
+
 - Add automatic conversion of ode models to linear models when
   detected.  This conversion is applied transparently at solve time
   (`rxSolve(..., useLinCmt=TRUE)`, the default) and the detected PK
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,16 +9,22 @@
   the original ODE model is used instead, so the conversion never
   breaks an otherwise-valid solve.
 
-- The automatic `linCmt()` conversion no longer linearizes a central
-  sub-system that is coupled to an additional `d/dt()` state which is
-  referenced elsewhere in the model (for example a peripheral or
-  metabolite observable such as `Cp <- periph / vp`, or a second
-  endpoint keyed to that state).  Because `linCmt()` exposes only the
-  central concentration, folding such states away dropped the coupled
-  state and demoted it to a required input parameter, so the default
-  solve aborted with "parameter(s) are required for solving:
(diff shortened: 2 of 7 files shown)


mattfidler:
In theory you should be able to fix this by renaming to the appropriate peripheral or central instead.

billdenney:
From Claude:

Thanks @mattfidler — you're right, the amounts are accessible and correct. I checked the values against a full 2-cmt ODE solve: `peripheral1` (amount) and `peripheral1/volume` (concentration) both match to 1e-3 in:

- ✅ classic `rxode2({...})` — both default (linCmtA) and `linCmtSens="linCmtB"`
- ✅ a UI `function(){}` model **without** an error model

There's one gap that's blocking it for the `odeToLin` / nlmixr2lib models, though: a UI model **with an error model (`~`)** treats the compartment name as an input parameter (= 0) rather than the linCmt amount.

```r
# correct — pAmt == true peripheral amount:
rxode2({ cl<-5; vc<-30; q<-2; vp<-50; Cc<-linCmt(); pAmt<-peripheral1 })

# broken — Cp solves to 0 (peripheral1 lands in params):
function() {
ini({ lcl<-log(5); lvc<-log(30); lq<-log(2); lvp<-log(50); propSd<-0.1 })
  model({
    cl<-exp(lcl); vc<-exp(lvc); q<-exp(lq); vp<-exp(lvp)
    Cc <- linCmt()
    Cp <- peripheral1/vp
    Cc ~ prop(propSd)
  })
}
```

It reproduces for single-endpoint, output-only, multi-endpoint (`Cp ~ ...`), and literal `linCmt() ~ ...` forms. It looks like the error-model path (`R/err.R` ~1362-1397) runs `_rxode2_linCmtGen`, which substitutes the `linCmt()` call but doesn't tie compartment-name references to the linCmt system, so they fall through to `params`. The non-error path compiles the body directly and the C parser materializes `peripheral1` (which is why classic notation works).

Is wiring that recognition through the error-model path the "UI transformation" you had in mind, or is there an existing hook I should use? I'll take a crack at it but want to make sure I match the intended design.

billdenney:
This now has a better fix by revised model rewriting. (Note that I updated the original PR title and description in case Matt's message above doesn't make sense anymore.)

mattfidler:
> t looks like the error-model path (R/err.R ~1362-1397) runs _rxode2_linCmtGen, which substitutes the linCmt() call but doesn't tie compartment-name references to the linCmt system, so they fall through to params. 

This is true, and a bug in the UI system.

mattfidler:
> Is wiring that recognition through the error-model path the "UI transformation" you had in mind, or is there an existing hook I should use? I'll take a crack at it but want to make sure I match the intended design.

This hook is described here:

https://nlmixr2.github.io/rxode2/articles/Integrating-User-Defined-Functions-into-rxode2.html#functions-to-insert-rxode2-code-into-the-current-model

The process is:

1. Create a `rxUdfUi.linCmt` s3 method.  If the first argument is a string and matches the possibilities above, convert it to the right value, that is `peripheral1/vp` or similar parameter depending on what the `linCmt()` algorithm picks up.  This only works when the `linCmt()` has one argument and that argument is specified as you were thinking above. 
2. Create a dummy `linCmt()` function that describes this new interface with roxygen2 as well as the ability to specify the parameters in the model directly, with a few examples.

This way the interface does not require an additional layer of parsing complexity and can be used in the ui/nlmixr2 model functions.

mattfidler:
Note that the following should also be false:

> a peripheral carrying its own estimated ~ endpoint is left as an ODE (it collides with linCmt's internal compartment).

There are tests that they will not collide, but they are in a very specific order.  If the compartments are named there is no issue here.  

The depot and central compartment numbers are before the ode compartment numbers.  The peripheral compartment are after the ode compartment numbers.  This is true regardless of the methods used but it should not be enough to stop it being modeled as an endpoint.  The issue is if the model parameters are mis-defined.  By evaluating the model using `m()` or similar you can see where the compartment numbers should be labeled.

PR #1092, opened 2026-06-22
Analytic-covariance building blocks: rxExpandSens3_ (3rd-order sensitivities) + rxOmegaVarCovDeriv (non-Cholesky Omega)

hiddevandebeek:
Two building blocks for analytic FOCEI/FOCE covariance matrices in nlmixr2est (nlmixr2/nlmixr2est#690), both finite-difference-free and validated against finite differences.

## 1. `rxExpandSens3_()` — analytic third-order forward sensitivities

Generates the **exact** third-order forward sensitivity equations `d^3(state) / d s1 d s2 d s3`, extending the existing first/second-order machinery (`rxExpandSens_()`, `rxExpandSens2_()`). `.rxSens()` gains a `vars3` argument that drives it and stores the result in `..sens3`.

Following `rxExpandSens2_`, the third-order RHS is the total derivative of the stored second-order RHS `v2 = d/dt(s^{s2,s3})` w.r.t. the new variable `s1`:

```
v3 = D(v2, s1)
   + sum_m [ D(v2, x_m)*s^{s1}_m  +  J*s^{s1,s2,s3}_m ]
   + sum_{q in unique{s2,s3}} sum_m D(v2, s^{q}_m)*s^{s1,q}_m
```

The `unique{s2,s3}` deduplication avoids a spurious factor of two when `s2 == s3`. `tests/testthat/test-sens3.R` checks the analytic third-order values against a central FD of the analytic second-order (~1e-6) and confirms symmetry.

These supply the structural sensitivities the **M3** term of the analytic FOCEI/FOCE covariance Hessian needs (`d^3f/d eta^3`, etc.).

## 2. `rxOmegaVarCovDeriv()` — non-Cholesky Omega derivatives

The default random-effects parameterization for estimation is a **Cholesky** decomposition, whose parameters are Cholesky factors rather than interpretable variances/covariances. Reporting random-effect SEs on the natural variance scale -- or assembling an analytic covariance over the `Omega` block -- needs a **non-Cholesky** path that differentiates w.r.t. the variance-covariance entries directly.

`rxOmegaVarCovDeriv(omega, order = 2)` returns `Omega^{-1}`, `log|Omega|`, and their first and second derivatives w.r.t. each free lower-triangular element `omega_{ab}`, in closed form:

```
d(Omega^-1)/d(omega_ab) = -Omega^-1 E_ab Omega^-1
d(log|Omega|)/d(omega_ab) = tr(Omega^-1 E_ab)
```

`tests/testthat/test-omega-varcov.R` validates the first and second derivatives against finite differences on a 3x3 **block** Omega (off-diagonals nonzero).

## Scope / blast radius

Both are additive: `rxExpandSens3_` is inert unless `.rxSens(..., vars3=)` is called; `rxOmegaVarCovDeriv` is a new standalone function. No parser, codegen, or solver changes.

## Checklist
- [x] `NEWS.md` entries
- [x] `rxTest({...})` tests (`test-sens3.R`, `test-omega-varcov.R`)
- [x] Bindings/exports: `rxExpandSens3_` (compileAttributes, init.c, NAMESPACE, man), `rxOmegaVarCovDeriv` (NAMESPACE, man)
- [x] Follows `.github/copilot-instructions.md` conventions


hiddevandebeek:
needed building block for nlmixr2/nlmixr2est#690

hiddevandebeek:
@mattfidler lmk what you think

mattfidler:
This looks good to me;  The sensitivity variables were already super long and these are even longer :smile: 

Please also add your name as a contributor to this package.

mattfidler:
As a general note an ODE of this size is more likely to fail solving than other methods.  

When adding to nlmixr2est, check for a failure in solve and fall back to the slow finite difference method.

mattfidler:
If you wish to do more, I was going to add a full covariance matrix in nlmixr2 for omegas, residuals; This would need to be handled in rxode2 as well since the default parameterization is a cholesky decomposition.  To get the SE there needs to be another path that has a non-cholesky decomposition omegas.

hiddevandebeek:
> This looks good to me; The sensitivity variables were already super long and these are even longer 😄
> 
> Please also add your name as a contributor to this package.

haha yes indeed. I will also add my name as contributor.

> When adding to nlmixr2est, check for a failure in solve and fall back to the slow finite difference method.

okay will do!

> If you wish to do more, I was going to add a full covariance matrix in nlmixr2 for omegas, residuals; This would need to be handled in rxode2 as well since the default parameterization is a cholesky decomposition. To get the SE there needs to be another path that has a non-cholesky decomposition omegas.

I have some time at the end of the week, I'll let you know if I manage to get to it!

hiddevandebeek:
As a note: the SE part for omegas was also where my previous attempt sort off got stuck and showed bias.

mattfidler:
Well the omega derivatives needed are clearly spelled out in the Almquist paper, the approch we use here is slightly different because of the transformation (if it helps).

hiddevandebeek:
yeah sorry I meant my original approach in nlmixr2/nlmixr2est#666 showed bias in the omegas

hiddevandebeek:
Pushed two more commits addressing the notes above:

**1. Added myself as a contributor** to `DESCRIPTION` 👍

**2. `rxOmegaVarCovDeriv()` — the non-Cholesky Ω path.** Since the default Ω parameterization is a Cholesky decomposition, getting Ω/residual SEs on the natural variance scale needs derivatives w.r.t. the variance-covariance elements directly. This returns `Omega^-1`, `log|Omega|`, and their first and second derivatives w.r.t. each free `omega_{ab}`, in closed form:

```
d(Omega^-1)/d(omega_ab)   = -Omega^-1 E_ab Omega^-1
d(log|Omega|)/d(omega_ab) =  tr(Omega^-1 E_ab)
```

Validated against finite differences (1st + 2nd derivatives) on a 3×3 **block** Ω in `tests/testthat/test-omega-varcov.R`. (These are the natural-scale forms; the existing Cholesky path is the transformed version you mentioned.)

On the nlmixr2est side, also done per your notes:
- **solve-failure → FD fallback**: the augmented sensitivity ODE is large, so every per-subject solve is guarded — any failure (or out-of-scope model) drops the whole population back to the finite-difference covariance, all-or-nothing.
- I prototyped the **full covariance (θ + Ω + residual)** using this non-Cholesky path and FD'd it in the variance-covariance scale. It matches NONMEM `$COVARIANCE MATRIX=R` (structural θ) and ferx across **all** blocks to ~1% — including the Ω and residual SEs. That's the spot where my gradient-only approach showed bias.

Happy to coordinate on the full-covariance-matrix feature whenever you'd like to pick it up.

mattfidler:
> As a note: the SE part for omegas was also where my previous attempt sort off got stuck and showed bias.

As a note, the omegas are not actually normally distributed, this may be where your assumptions fell apart with the bias.

mattfidler:
This is awesome, I will look further into it soon.

As I think about the robustness of the Hessian, there is a missing fallback:

- 3rd order exact, if it solves
- Missing 2nd order with finite differences of the 2nd order gradients (These should use shi differences gated on the harmonic mean), if it solves
- Full Finite difference

The accuracy and speed would be best for the first case, then second, then third.

I'm OK taking on the second if you want, since the documentation of what was needed and how it was done is lost in a poster somewhere (not actually published).

hiddevandebeek:
@mattfidler perfect, ill make the fallback logic a bit more sound!

I sent you an email btw for a more rigorous proof and possible collab on this.

hiddevandebeek:
the FD of the second order is indeed a nice addition and makes a lot of sense.

mattfidler:
The failure is from not fixing documentation.  Can you fix this so the CI will return success.

mattfidler:
Also, while I appreciate the `@author` tag, it really shouldn't be me.

hiddevandebeek:
Before I commit to porting the nlmixr2est covariance to C++, I wanted to check you're happy with the **theory** behind the two primitives this PR adds, since they're what it's built on:

- `rxExpandSens3_` — 3rd-order forward sensitivities, needed for the covariance's log-determinant term (the inner mode moves with `θ`);
- `rxOmegaVarCovDeriv` — the non-Cholesky Omega variance–covariance derivatives (`∂Ω⁻¹/∂ω_ab = −Ω⁻¹ E_ab Ω⁻¹`, plus the `log|Ω|` and 2nd-derivative terms), so the covariance comes out on the natural variance/block-covariance scale rather than Cholesky.

Both are validated through the full covariance in nlmixr2est (matches NONMEM `$COV MATRIX=R` / ferx ~1%, diagonal and block Omega). **Do these look like the right primitives / is the approach sound to you?** If so, the covariance moves to C++, and I'd expect `rxOmegaVarCovDeriv` (currently R) to move alongside it.

mattfidler:
This looks good to me, since you want to move this to C/C++ feel free. Let me know when you are ready.

mattfidler:
If you are calling the C from nlmixr2est, you will need to add it to the rxode2 api so it can be called in parallel when necessary.

hiddevandebeek:
@mattfidler okay ported to C++, this is part in rxode2 is done one CMD checks clear. I tried to follow code convention.

btw did you see my email, since i got an auto-reply that you're out of office? If you want to respond later that's also fine!

mattfidler:
I got your email but since I'm out of office it is harder for me to look at the document on my phone.

I plan to review when I get back in the office next week.

Matt

hiddevandebeek:
ah no hurry, thought I'd ask since I saw you were quite busy here!

mattfidler:
Yes, but I have a different computer where I have word, and am looking at this while waiting for other things :smile: 

A distracted review of math doesn't quite work for me.

This is not exactly the API that I was referring to, however I don't think this will be called in an OpenMP region so it is likely not an issue.

With that in mind, I am merging the request.

PR #1093, opened 2026-06-24
Dde dense

mattfidler:


mattfidler:
@billdenney 

This  is the dde interface you were wanting.

billdenney:
Thank you! Looking at the documentation, can it only be for states or can it also be for arbitrary math. For example, would the following work:

```
function() {
  ini({
    kel = 1
    v = 2
  })
  model({
    d/dt(central) <- -kel*central
    conc <- central/v
    concDelay <- delay(conc, 5)
  })
}
```

mattfidler:
Only works for states, just like monolix and nonmem.  That model would error and say it is invalid.

```
rxode2 model syntax error:
================================================================================
:001: d/dt(central) <- -kel * central
:002: conc <- central/v
:003: 'delay' can only be used on a 'd/dt()' line:
      concDelay <- delay(conc, 5)
                   ^~~~~
================================================================================
Error: syntax errors (see above)
```

PR #1106, opened 2026-07-09
fix(dde): correct delay history when a delay model's Jacobian can't be generated (+ enable ros4 for delayed sensitivity models)

mattfidler:
## Summary

Two related fixes for delay differential equation (`delay()`) models that request an implicit method (`ros4` or the `dop853+ros4` composite). Together they fix a silent-wrong-results bug and enable the stiff `ros4` leg for delayed sensitivity models (e.g. nlmixr2est's analytic-covariance augmented models).

### 1. Fall back to `dop853` (dense), not `liblsoda`, when the Jacobian can't be generated

A `delay()` model whose analytic Jacobian generation failed fell back to `liblsoda`. `liblsoda` is **non-dense and records no delay history**, so every `delay()` lookup silently returned the pre-history `past()`/initial-condition value and the solution diverged (no error, just wrong numbers). A delay model requires a dense solver; `dop853` dense needs no Jacobian, so fall back to it instead.

### 2. Keep the `d/dt()` of delayed sensitivity-named states so `calcJac` works

*Why* Jacobian generation failed above: a delayed state whose name matches the sensitivity convention (`rx__sens_<state>_BY_<var>__`) had its `d/dt()` reclassified as a derivable sensitivity and stripped -- in both the `calcJac`/`calcSens` model rewrite (`rxGetModel`) and the symengine load (`rxS`) -- leaving a `delay()` with no defining ODE (`delay(X, ...) has no d/dt(X)`). A state read by `delay()` is a genuine ODE (`delay(X, T)` requires `d/dt(X)`), so its `d/dt()` is now always kept:

- `rxGetModel` excludes delayed states from the sensitivity strip (new `.rxSensStrippable()`).
- `rxS` records `..delayedStates` and routes their `d/dt` to `..ddt..` instead of `..sens0..`.

With this, these models generate a Jacobian and use the stiff `ros4` leg directly rather than falling back.

## Verification

- `ros4` / `dop853+ros4` on the augmented sensitivity model now match pure `dop853` to ~2e-10 (previously diverged ~0.28).
- `ros4` genuinely runs (no `liblsoda` fallback) and matches the dense reference to ~5e-7 -- confirming ros4's cubic delay-history reconstruction is itself exact; the divergence was entirely the fallback path.
- `test-dde`, `test-dde-past`, `test-dde-sens`: all green, no regressions.
- Changes are guarded by the model's `hasDelay` flag, so non-delay Jacobian/sensitivity generation is unchanged.
- Downstream nlmixr2est DDE FOCEI fit unchanged (analytic covariance, all SEs finite).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,12 @@
 # rxode2 5.1.3
 
+- Fix delay (`delay()`) models that request an implicit method (`ros4` or the
+  `dop853+ros4` composite) but whose analytic Jacobian cannot be generated:
+  they previously fell back to `liblsoda`, which is non-dense and records no
+  delay history, so `delay()` lookups silently returned the pre-history
+  `past()`/initial-condition value and the solution diverged.  Such models now
+  fall back to `dop853` (dense, no Jacobian required) instead.
+
 - Fix `lag()`/`diff()` (and `first()`/`last()`) which previously returned a
   constant instead of the previous record's value, for both calculated (lhs)
   variables and time-varying covariates. They now read the prior record's value
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,13 @@
 # rxode2 5.1.3
 
+- Fix analytic Jacobian generation (`calcJac=TRUE`) for models whose delayed
+  states use the sensitivity naming convention (`rx__sens_<state>_BY_<var>__`,
+  e.g. nlmixr2est's analytic-covariance augmented models): the `d/dt()` of such
+  a state was reclassified as a derivable sensitivity and stripped, leaving a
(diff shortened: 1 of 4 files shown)


mattfidler:
Companion to nlmixr2/nlmixr2est#703 (FOCEI analytic covariance + DDE fast-family support). That branch's DDE augmented-covariance path depends on this fix; its `.foceiAnalyticSolveFA`/`.foceiAnalyticSolveAll` currently force `method=dop853` as a belt-and-suspenders workaround, and `tests/testthat/test-dde-focei.R` there guards the end-to-end pipeline.

Issue #1119, opened 2026-07-12
eventSens jump sensitivities are zero/absent for ODE + linCmt() models


mattfidler:
## Summary

The analytic `eventSens="jump"` event-sensitivities (for modeled event-timing parameters like `alag`/`f`) are broken or absent in models that contain `linCmt()`. Found while verifying #1118 (the coincident-time jump fix).

## Reproducer

A modeled lag on an ODE compartment behind a `linCmt()` term yields a **zero** sensitivity everywhere, while the identical pure-ODE model is correct:

```r
# pure ODE: rx__sens_peri_BY_eta_lag__ is correct (nonzero, matches FD)
mOde <- rxode2({ ka<-exp(tka); alag(depot)<-2*exp(eta_lag)
  d/dt(depot)<--ka*depot; d/dt(peri)<-ka*depot-0.3*peri }, calcSens="eta_lag", eventSens="jump")

# add a linCmt term: rx__sens_peri_BY_eta_lag__ becomes 0 at every time
mMix <- rxode2({ ka<-exp(tka); alag(depot)<-2*exp(eta_lag)
  d/dt(depot)<--ka*depot; d/dt(peri)<-ka*depot-0.3*peri
  C2<-linCmt(CL,V) }, calcSens="eta_lag", eventSens="fd",
  linCmtSens="linCmtA", linCmtSensType="A")
```

Separately, **pure** linCmt models expose only structural (`p1`/`v1`/`ka`) sensitivities; a modeled-lag/dose parameter on a linCmt-fed compartment produces no sensitivity at all.

## Root causes

1. **Reserved-name collision** — `.rxEventSensFilterMap` computes ODE states via `setdiff(mv$normal.state, .rxLinCmt(mv))`, but `.rxLinCmt` returns hard-coded generic names (`depot`, `central`, `peripheral1`, ...). A physical ODE state named `depot`/`central` is wrongly dropped, collapsing the jump map and emptying `lagCmt`, so the jump never fires. (Non-reserved names like `gut`/`eff` work today.)

[...]

PR #1120, opened 2026-07-12
eventSens: linCmt() models downgrade to finite differences + reserved-name guard

mattfidler:
Resolves the linCmt() event-sensitivity issues investigated in #1119 (follow-up to the merged coincident-time jump fix #1118).

## What this does

**Net change is R-only** (`R/eventSens.R`, `R/rxode2.R`, tests):

1. **Downgrade linCmt() models to finite differences** for event-timing (modeled `alag`/`f`) sensitivities. The analytic moving-boundary (jump) sensitivity for a linCmt compartment is not implemented; rather than emit an incomplete/incorrect analytic result, all linCmt models (`numLin > 0`) resolve `eventSens` to `"fd"` and use the FOCEi/nlmixr2est finite-difference event path. Structural-parameter linCmt sensitivities are unaffected (continuous, handled by `linCmtB`). — `.rxEventSensEffectiveMode()`.

2. **Reserved-name collision guard.** Naming an ODE compartment `depot`/`central`/`peripheralN` alongside an oral/multi-cmt `linCmt()` conflates the ODE and linCmt compartments, so the ODE state loses its sensitivity expansion and its sensitivities are silently incorrect. This now warns at build time (`rxode2()`), via the extracted `.rxLinCmtNameCollision()` helper, so it fires even under the FD downgrade.

3. **`.rxLinCmtEventSensPairs()`** — detects `(linCmt compartment, calcSens param)` pairs with a modeled `alag`/`f`; kept for a future analytic implementation.

## Background

Investigation (#1119) established that the originally-reported "mixed ODE+linCmt jump = 0" was a malformed model (an ODE compartment named `depot` colliding with an oral linCmt's `depot`), and that linCmt sensitivities materialize as chain-ruled `linCmtB(which=-2,j)` LHS assignments in the FOCEi inner model — not as solve-compartments. A proper analytic linCmt moving-boundary sensitivity is a larger cross-package (rxode2 + nlmixr2est) effort tracked in #1119; this PR ships the safe FD downgrade + the collision guard in the meantime.

## Testing

`devtools::test()` on the relevant suites (via `load_all`, not install): event-sensitivities **185 pass / 0 fail**, lincmt-solve **1569 pass / 0 fail**.

## Note on history

The branch contains intermediate WIP commits that explored a solve-compartment materialization; the final commit reverts them (superseded by the LHS mechanism), so the **net diff is clean and R-only**. Squash-merge recommended.

https://claude.ai/code/session_01JwVDp22R3CgEP22sXJB7ZU

### DIFF
--- a/R/eventSens.R
+++ b/R/eventSens.R
@@ -333,6 +333,48 @@
   list(lagCmt = .lag, fCmt = .bit(2L), rateCmt = .bit(8L), durCmt = .bit(16L))
 }
 
+#' Identify (linCmt compartment, event-timing parameter) pairs needing sensitivities
+#'
+#' For a linCmt() model, a modeled alag()/f() on a linCmt compartment driven by a
+#' `calcSens` parameter has an event-timing (moving-boundary) sensitivity that the
+#' structural linCmt Jacobian (wrt p1/v1/ka/...) does not capture.  This returns
+#' the (linCmt state, driving parameter) pairs so the linCmt solve can carry the
+#' extra sensitivity columns (Part B of nlmixr2/rxode2#1119).
+#'
+#' @param obj Built model (rxUi, rxode2, modelVars).
+#' @param calcSens Character vector of first-order sensitivity parameters.
+#' @return data.frame(state, param, kind, cmt) (1-based `cmt`), or `NULL` when
+#'   the model has no linCmt event-timing sensitivities.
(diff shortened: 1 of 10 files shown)


mattfidler:
Superseded by #1121 (same fixes, clean single-commit history off origin/main).

PR #1125, opened 2026-07-13
linCmtB(): thread-safe forward-mode AD Jacobian solves

mattfidler:
## Summary

Makes `linCmt()` sensitivity solves (`linCmtB`) thread safe on the default
forward-mode AD Jacobian path, so parallel subject solving is no longer forced
to a single core for these models.

Previously any model containing `linCmtB` was flagged `notThreadLinCmtB` at
parse time and forced to `op->cores = 1` at solve time -- a guard dating from
when the Jacobian used Stan's shared reverse-mode arena. The default path is now
forward-mode `fvar` (stack-local, no shared arena), so it is safe to run per
subject across threads.

## Changes

- **Enable threading on the AD path** (`src/rxData.cpp`): the
  `notThreadLinCmtB` solve-time cases now thread like any thread-safe method
  when `sensType` is forward-mode AD (`3`/`30`/auto `100`). Reverse-mode AD
  (`31`, shared `ChainableStack`) and the finite-difference paths (which share a
  first-subject scaling/step setup) stay single threaded.
- **Fix a cross-thread race** (`src/linCmt.cpp`): `linCmtScaleInitPar`/`N` and
  `linCmtZeroJac` read the hardcoded `__linCmtB[0]` slot, but the
  finite-difference H setup calls them *after* `ind_linCmtFH` has populated the
  *current thread's* slot -- so under threading they raced (and could read a
  mid-resize slot). They now read `__linCmtB[rx_get_thread()]`.
- **Skip finite-difference H setup on the AD paths** (`src/par_solve.cpp`):
  `setupLinH()` ran `shi21ForwardH()` (a base solve plus per-parameter probe
  solves) for the default AD sensType even though the AD Jacobian never reads
  `ind->linH`. `linCmtSensIsAD()` is given external linkage (its FD set is
  exactly the sensTypes `linCmtB` reads `ind->linH` for) and `setupLinH()`
  early-outs for AD; the dead `case 3` branch is removed. Output is unchanged;
  this removes redundant work on the FOCEi (`ind_solve`) path.

## Validation

- 80-subject `linCmtB` sensitivity solve is **bitwise identical** at `cores=1`
  vs `cores=4`; reverse-AD (`"ADr"`) stays serial and correct. New regression
  test in `test-lincmt-solve-sens.R`.
- AD `linCmtB` vs ODE reference unchanged (3.5e-7 < 1e-5 tol); FOCEi `theo_sd`
  fit gives identical OFV before/after the H-setup change.
- `lincmt-solve` (1623), `lincmt-solve-sens` (9), `lincmt-cmt-ref` (54),
  `lincmt-parse` (4580) all pass.

Note: the per-subject FOCEi *inner-loop* parallelization that most benefits from
this lives in `nlmixr2est`; this PR makes rxode2's side (a thread-safe
per-subject `linCmtB`) ready for it.


mattfidler:
Superseded by a clean PR based directly on main (this branch accidentally included unrelated par-loader commits from a local main that was ahead of origin/main).

PR #1126, opened 2026-07-13
linCmtB(): thread-safe forward-mode AD Jacobian solves

mattfidler:
## Summary

Makes `linCmt()` sensitivity solves (`linCmtB`) thread safe on the default
forward-mode AD Jacobian path, so parallel subject solving is no longer forced
to a single core for these models.

Previously any model containing `linCmtB` was flagged `notThreadLinCmtB` at
parse time and forced to `op->cores = 1` at solve time -- a guard dating from
when the Jacobian used Stan's shared reverse-mode arena. The default path is now
forward-mode `fvar` (stack-local, no shared arena), so it is safe to run per
subject across threads.

## Changes

- **Enable threading on the AD path** (`src/rxData.cpp`): the
  `notThreadLinCmtB` solve-time cases now thread like any thread-safe method
  when `sensType` is forward-mode AD (`3`/`30`/auto `100`). Reverse-mode AD
  (`31`, shared `ChainableStack`) and the finite-difference paths (which share a
  first-subject scaling/step setup) stay single threaded.
- **Fix a cross-thread race** (`src/linCmt.cpp`): `linCmtScaleInitPar`/`N` and
  `linCmtZeroJac` read the hardcoded `__linCmtB[0]` slot, but the
  finite-difference H setup calls them *after* `ind_linCmtFH` has populated the
  *current thread's* slot -- so under threading they raced (and could read a
  mid-resize slot). They now read `__linCmtB[rx_get_thread()]`.
- **Skip finite-difference H setup on the AD paths** (`src/par_solve.cpp`):
  `setupLinH()` ran `shi21ForwardH()` (a base solve plus per-parameter probe
  solves) for the default AD sensType even though the AD Jacobian never reads
  `ind->linH`. `linCmtSensIsAD()` is given external linkage (its FD set is
  exactly the sensTypes `linCmtB` reads `ind->linH` for) and `setupLinH()`
  early-outs for AD; the dead `case 3` branch is removed. Output is unchanged;
  this removes redundant work on the FOCEi (`ind_solve`) path.

## Validation

- 80-subject `linCmtB` sensitivity solve is **bitwise identical** at `cores=1`
  vs `cores=4`; reverse-AD (`"ADr"`) stays serial and correct. New regression
  test in `test-lincmt-solve-sens.R`.
- AD `linCmtB` vs ODE reference unchanged (3.5e-7 < 1e-5 tol); FOCEi `theo_sd`
  fit gives identical OFV before/after the H-setup change.
- `lincmt-solve` (1623), `lincmt-solve-sens` (9), `lincmt-cmt-ref` (54),
  `lincmt-parse` (4580) all pass.

Note: the per-subject FOCEi *inner-loop* parallelization that most benefits from
this lives in `nlmixr2est`, which already clamps its inner core count to
rxode2's `op->cores` -- so this PR makes that parallelization kick in for
`linCmt()` models automatically.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -78,6 +78,11 @@
   inside `linCmt()` and mixed `linCmt()`+ODE models, with Jacobian handling of
   the dosing events; `odeToLin()` preserves and renames them when converting.
 
+- `linCmt()` sensitivity (`linCmtB`) solves now run in parallel across subjects
+  on the default forward-mode AD Jacobian path (`linCmtSensType="AD"`), which is
+  stack-local with no shared Stan arena.  The reverse-mode AD (`"ADr"`) and
+  finite-difference paths remain single threaded.
+
 - Inductive linearization and matrix exponentials rewritten with a more
   NONMEM-like interface (automatic ODE->syntax translation retained) and
   symbolic-differentiation gradients.
--- a/src/linCmt.cpp
+++ b/src/linCmt.cpp
@@ -493,8 +493,11 @@ extern "C" int linCmtZeroJac(int i) {
 // True for the automatic-differentiation jacobian methods (forward-mode
 // fvar = 3/30, reverse-mode = 31, and the auto/default path), which all want
 // the unscaled thetaSens + passthrough trueTheta. The finite-difference
-// methods (1,2,4,5,6,7,10,20,40,50) keep the scaled path.
-static inline bool linCmtSensIsAD(int sensType) {
+// methods (1,2,4,5,6,7,10,20,40,50) keep the scaled path.  These same FD
+// methods are exactly the ones linCmtB reads ind->linH for, so setupLinH
+// (par_solve.cpp) uses this predicate to skip step-size estimation on the AD
+// paths -- external linkage lets it share this single source of truth.
+bool linCmtSensIsAD(int sensType) {
   switch (sensType) {
   case 1: case 2: case 4: case 5: case 6: case 7:
   case 10: case 20: case 40: case 50:
--- a/src/linCmt.cpp
+++ b/src/linCmt.cpp
@@ -8,6 +8,7 @@
 #include "timsort.h"
 #define SORT gfx::timsort
 #include "linCmt.h"
+#include "linCmtSensType.h"
 
 
 extern rx_solving_options op_global;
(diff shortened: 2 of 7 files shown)


mattfidler:
## Validated end-to-end in nlmixr2est FOCEi

Confirmed that this rxode2 change alone enables parallel FOCEi for `linCmt()`
models -- no nlmixr2est change is needed. nlmixr2est's `innerOpt()` already sets
its inner-loop core count from rxode2's `op->cores` and gates on
`solveMethodThreadSafe(op)`; with the linCmtB thread-safe gate here, the pool
setup solve reports `op->cores=4` and the inner subject loop runs with
`_doParallel=1`.

Timing, 192-subject 1-cmt oral `linCmt()` FOCEi fit (theo_sd x16):

```
cores=1: 11.23s  OFV=1868.8616
cores=4:  4.75s  OFV=1868.8616   -> 2.36x, OFV delta 0.00e+00
```

Identical objective at both core counts.

mattfidler:
Thanks -- addressed the Copilot review in 5ad561e23.

**`sensType == 100` (auto) treated as thread-safe (rxData.cpp)** -- verified `auto` always resolves to forward-mode AD, so this is safe, and made the reasoning explicit. The only runtime mutation of `rx->sensType` is `setupLinH()`'s `100 -> 3` remap (both forward-mode AD); no code path resolves `auto` to reverse-mode AD (31) or a finite-difference method (R sets `linCmtSensType` only from the user arg, no feature-based downgrade -- the `eventSens` FD downgrade is a separate axis and leaves the linCmtB Jacobian method untouched). The gate now calls a named `linCmtSensForwardAdThreadSafe()` predicate whose comment documents why `100` is included and why reverse-mode AD and the FD paths are excluded.

**Cross-TU forward declaration of `linCmtSensIsAD` (par_solve.cpp)** -- moved both classifiers into a shared header `src/linCmtSensType.h` (`static inline`), included by linCmt.cpp, par_solve.cpp and rxData.cpp. Removed the forward declaration and the external definition; single source of truth, no drift.

**`__linCmtB.size()` narrowing (linCmt.cpp)** -- `rx_get_thread()` takes an `int`, so each accessor (`linCmtScaleInitPar`/`linCmtScaleInitN`/`linCmtZeroJac`) now casts the `size_t` pool size to `int` once and stores the thread index in a local.

**Test could pass vacuously if clamped to 1 core (test-lincmt-solve-sens.R)** -- rxode2 exposes no per-solve "cores actually used" telemetry, so the test now `skip_if_not(getRxThreads() >= 2L)`; on a multi-thread runner the `cores = 2` request genuinely exercises the parallel path (this model's `linCmtB` carries the forward-AD flag the gate now honors), and it skips rather than passing trivially otherwise.

Validated: clean rebuild with no narrowing/unused-function warnings; `lincmt-solve-sens` 9 PASS (the thread test ran, not skipped) and `lincmt-solve` 1623 PASS.

PR #1145, opened 2026-07-18
Remove $predDf ar column; parse error-model literals to FIX $iniDf params

mattfidler:
## Why

`$predDf` and `$iniDf` are a **stable, exported interface** relied on by reverse dependencies (nlmixr2est, babelmixr2, nonmem2rx, …) and by fits/UIs saved with earlier versions. The recently-added `ar` column (for the `ar()` AR(1) residual term) broke that contract:

- The released CRAN **nlmixr2est** and previously-stored UIs lack the column, so `.rxGetArCorLang()`'s `is.na(pred1$ar)` errored with *"argument is of length zero"*.
- Stale/older fits failed in `predict()` and `saemModelPred`.

Since a released reverse dependency cannot be patched retroactively, the fix belongs here in rxode2.

## What changed

- **Error-model literals → FIX `$iniDf` params.** A numeric literal in any error function (`add(0.7)`, `prop(0.1)`, `ar(0.5)`) is now parsed into an auto-generated, uniquely named, `rx`-prefixed **FIX** parameter (`rx.<endpoint>.<func>`), keyed to its endpoint via the existing `err`/`condition` columns — exactly like a user-written fixed residual parameter (`.rxErrLiteralToFixParam`). This also fixes literals like `add(0.7)` that were previously **silently dropped**.
- **Removed the `ar` column from `$predDf`** and updated every consumer (`.rxGetArCorLang`, `rxHasAr`, `assertRxUiEstimatedResiduals`, the mu-ref parameter check).
- **All three `ar()` forms keep working:** literal → FIX param; estimated `ar(rho)` → keeps its `err=="ar"` iniDf linkage; modeled `ar(corv)` (a calculated variable) → recovered directly from the endpoint's error expression (`.rxFindArArg`), with `rxHasAr()` scanning error lines to detect it — no new column needed.
- **CLAUDE.md guardrail:** `$predDf`/`$iniDf` schema changes require explicit permission; reverse-dependency compat fixes belong in rxode2, not in the reverse dependency.
- NEWS entry + updated `test-ar.R`.

## Verification

- rxode2 (`load_all`): `test-ar` (34), `test-ui-err` (229), `test-rxFix`, `test-parsing`, and a 174-test sim/focei/deparse regression sweep — **all pass, 0 failures**.
- Reverse-dep: **unpatched** nlmixr2est against the rebuilt rxode2 — `predict` (11), `issue-281` (14), `issue-502` (1), `keep` (96) — all pass.

Note: the original `packageVersion("rxode2")` *"no package called 'rxode2'"* errors were a `load_all`-without-install artifact, not a real rxode2 break — they do not occur when rxode2 is installed (the real reverse-dep/CRAN scenario).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -124,6 +124,23 @@ devtools::document()
 - `rxUiGet` is an S3 dispatch system: `ui$property` calls `rxUiGet.property(x)`
 - Model piping (`R/piping*.R`) copies and modifies `rxUi` objects via `.copyUi()`
 
+> [!IMPORTANT]
+> **`$predDf` and `$iniDf` are a stable, exported interface — do not change
+> their schema without explicit user permission.** Their columns and structure
+> are relied on for model export/serialization and by reverse dependencies
+> (nlmixr2est, babelmixr2, nonmem2rx, …), including fits/UIs that were saved by
+> earlier versions. Adding, removing, renaming, or repurposing a column is a
+> breaking change: never do it as an incidental part of another change — ask
+> first. Prefer representing new information with the *existing* schema (e.g. a
+> literal residual value becomes an auto-generated FIX row in `$iniDf`, keyed to
+> its endpoint via the existing `err`/`condition` columns — see
+> `.rxErrLiteralToFixParam` in `R/err.R` — rather than a new `$predDf` column).
+>
+> **Reverse-dependency compatibility is fixed here, in rxode2.** When an rxode2
+> change breaks a reverse dependency, the released reverse dependency cannot be
+> patched retroactively, so the fix belongs in rxode2 — not in the reverse
+> dependency.
+
 **Event Table** (`src/et.cpp`, `src/etTran.cpp`, `R/et.R`):
 
 - `et()` constructs dosing/sampling event tables
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,20 @@
 # rxode2 5.1.3
 
+## Breaking / compatibility changes
+
+- The `ar` column that had been added to `$predDf` was removed.  `$predDf` and
+  `$iniDf` are a stable, exported interface relied on by reverse dependencies
+  (nlmixr2est, babelmixr2, ...) and by fits saved with earlier versions, so
+  their schema must not grow.  Instead, a numeric literal supplied to an
+  error-model function (e.g. `add(0.7)`, `prop(0.1)`, `ar(0.5)`) is now parsed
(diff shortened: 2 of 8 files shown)


mattfidler:
All six Copilot review comments are addressed in c2dd5f3e7:

1. **`.rxErrLiteralToFixParam` de-dup suffix** (`R/err.R`) -- counter now starts at `0L`, so the first collision yields `rx.<endpoint>.<func>.1` instead of `.2`.
2. **`ar()` literal FIX bound** (`R/err.R`) -- upper bound is now the strict-below-1 sup (`1 - .Machine$double.eps`) to match the `[0, 1)` domain; comment updated to match.
3. **`.rxFindArArg` LHS unwrapping** (`R/err-sim.R`) -- now unwraps `ll(x)` -> `x` and `linCmt()` -> `rxLinCmt` on the tilde LHS, mirroring `.errHandleLlOrLinCmt`, so modeled `ar()` correlations are recovered for `ll()`/`linCmt()` endpoints. Verified all three LHS forms.
4. **Test coverage for non-`ar()` literals** (`tests/testthat/test-ar.R`) -- added assertions that `add(0.7)`/`prop(0.1)` literals produce `rx.<endpoint>.<func>` FIX rows (est/err/condition/lower) and are not dropped.
5. **`cran-comments.md` grammar** -- "caused by or new to" -> "caused by nor new to".
6. **`cran-comments.md` hyphenation** -- "un-initialized" -> "uninitialized".

Verification (`load_all`): `test-ar` (45), `test-ui-err` (229), `test-rxFix` (7) -- all pass, 0 failures.

mattfidler:
Follow-up in 3e434bf08 -- multi-argument fixed error literals:

- **Naming fix (`.rxErrLiteralToFixParam`, `R/err.R`)** -- the auto-generated FIX parameter name now mirrors the endpoint's `err` specification (`funName` for argument 1, `funName<argno>` otherwise) instead of the bare function name. This means a multi-argument error function with more than one fixed literal (e.g. `pow(1.5, 2)`) produces distinct names keyed to their `err` rows (`rx.cp.pow`, `rx.cp.pow2`) rather than colliding on `rx.cp.pow` and falling back to a numeric dedup suffix.
- **Tests (`test-ar.R`)** -- added coverage for `pow(alpha, 1.5)` (estimated first arg stays a normal param; literal exponent fixed as `rx.cp.pow2`, `err == "pow2"`) and `pow(1.5, 2)` (both literals fixed to `rx.cp.pow` / `rx.cp.pow2`).

Verification (`load_all`): `test-ar` (60), `test-ui-err` (229) -- all pass, 0 failures.

PR #1148, opened 2026-07-18
Fix CRAN R-devel NOTEs: deprecated structure() names and Rd HTML validation

mattfidler:
Fixes the two NOTEs from the R-devel Windows CRAN check of rxode2 5.1.3.

## NOTE 1 - deprecated `structure()` special names
`checking R code for possible problems` flagged calls to `structure()` using
deprecated special argument names. Renamed to the current forms:

- `.Label` -> `levels`, `.Names` -> `names`, `.Dim` -> `dim`, `.Dimnames` -> `dimnames`
- `R/err.R`: 6 factor-building `structure()` calls (`.Label` -> `levels`)
- Test files: `test-100-cmt`, `test-backward`, `test-etTrans`, `test-example-3-2`,
  `test-example-3-3`, `test-ini`, `test-named-id`, `test-occ`,
  `test-rxode-issue-007`, `test-ui-mv`

## NOTE 2 - HTML validation problem
`checking HTML version of manual` reported `<state> is not recognized` in
`dot-rxPastBaseLinesFromEnv.Rd`. The roxygen text `rx__pastRhs_<state>__`
emitted a literal `<state>` HTML tag. Rewrote the `R/dde.R` source with
backticked `rx__pastRhs_STATE__` / `rx__pastTau_STATE__` (no angle brackets)
and regenerated the man page via `devtools::document()`.

## Verification
- All 10 affected test files pass (714 tests, 0 failures) via `devtools::test()`.
- Verified the `err.R` factor builders still produce valid `errType`/`addProp`/
  `transform` factors with correct levels.
- Confirmed the regenerated `dot-rxPastBaseLinesFromEnv.Rd` no longer contains a
  raw `<state>` tag.


mattfidler:
Follow-up commit (49e06fa20): replaced the install-time `reexports.Rd` `\value` patch in `inst/tools/workaround.R` with a proper roxygen `@return` on a documented `reexports` name block in `R/reexport.R`. roxygen now emits `\value{Inherited from parent routine}` natively and idempotently across `devtools::document()` runs, so the hand-patch is no longer needed and won't get stripped/re-added each build.

PR #1153, opened 2026-07-23
fix(linCmt): reset per-subject linCmt() amounts in iniSubject (batched multi-subject alag leak)

hiddevandebeek:
## Summary

A batched multi-subject solve of a `linCmt()` model returns **wrong predictions for every subject after the first one on each OpenMP thread**. The result is correct only when `cores >= nSub` (each subject gets its own thread):

- `cores = 1` → all-but-subject-1 wrong
- `cores = 2` → about half wrong
- `cores >= nSub` → correct

Pure-ODE models are unaffected, and so is `linCmt()` with `f()`/`rate()`/`dur()`. A **modeled `alag()`/`lag()`** is what surfaces the bug (it relocates the dose in time, so stale amounts are read into the pre-dose window). This does not affect ordinary `nlmixr2` estimation, which converts `linCmt()` to ODEs (`linToOde`) before fitting; it surfaces when the native `linCmt()` path is driven directly over many subjects (e.g. population simulation).

## Root cause

`iniSubject()` re-initialises a subject's ODE state from `op->inits`, but only when `u_inis != NULL`. A **pure `linCmt()` model has no ODE states**, so `u_inis == NULL` and neither `memcpy` branch runs, leaving two `linCmt()` inter-event amount ("alast") buffers uncleared between subjects:

1. **`ind->linCmtSave`** — the **per-thread** save buffer (`gLinSave` sliced by thread id in `getLinCmtSaveThread()`), shared by every subject that runs on the same thread, otherwise never reset. Subject *k>1* on a thread therefore starts from subject *k-1*'s amounts. **This is the cross-subject leak.**
2. **`getAdvan(0) = ind->solve + op->linOffset`** — slot 0 of the per-subject advance buffer, which the fast-path `ind_linCmt0()` reads at `solvedIdx == 0` before the (possibly `alag`-relocated) dose re-initialises it.

## Fix

Zero both buffers in `iniSubject()` for models with `linCmt()` compartments (`op->numLin + op->numLinSens > 0`). `linCmt()` initial conditions are always zero, so this is exact. ODE/mixed models (`u_inis != NULL`, whose `memcpy(op->inits)` already covers the linCmt region) and pure-ODE models (`numLin == numLinSens == 0`) are untouched, so **ODE solves are byte-for-byte unchanged**. One file, `src/par_solve.h`.

## Reproduction

```r
library(rxode2)
m    <- rxode2("cl=exp(tcl+eta.cl); v=exp(tv); ka=exp(tka); alag(depot)=exp(tlag); cp=linCmt()")
tms  <- c(0.9,1.7,3,6,10)
ev   <- et(et(amt=100, cmt="depot"), tms)
th   <- c(tcl=log(3), tv=log(30), tka=log(1), tlag=log(0.7))
etas <- c(-0.5, 0, 0.5)

# reference: each subject in its own solve
ref <- t(sapply(etas, function(e)
  rxSolve(m, params=c(th, eta.cl=e), events=ev, returnType="data.frame", addDosing=FALSE)$cp))

df <- data.frame(tcl=th[["tcl"]], tv=th[["tv"]], tka=th[["tka"]], tlag=th[["tlag"]], eta.cl=etas)
for (cr in c(1L, length(etas))) {
  o  <- rxSolve(m, params=df, events=ev, cores=cr, returnType="data.frame", addDosing=FALSE)
  cm <- matrix(o$cp[o$time %in% tms], 3, 5, byrow=TRUE)
  cat(sprintf("cores=%d  max|batched-ref| = %.4g\n", cr, max(abs(cm-ref))))
}
```

| | `cores=1` | `cores=2` | `cores=3` |
|---|---|---|---|
| before | `max\|diff\| = 2.184` | `1.953` | `0` |
| after  | `0` | `0` | `0` |

The stale contribution decays at exactly `ke = cl/v`, confirming it is a leaked central-compartment amount from the previous subject on the thread.

## Testing

Built from source and verified in a dedicated session. In addition to the reproduction above (now correct at every `cores`), the following all remain correct at `cores=1` batched over multiple subjects (no regression): plain `linCmt()` without lag; `linCmt() + f(depot)`; `linCmt() + rate(depot)`; a pure ODE model **with** `alag()`; and the single-subject `linCmt()+alag()` case. Also verified via an id-keyed multi-subject event table.


mattfidler:
Thanks @hiddevandebeek

PR #1159, opened 2026-07-24
fix(dsl): centre the smoothed derivative of relational operators on the discontinuity

hiddevandebeek:
## Summary

The symbolic derivatives of the relational operators (`>`, `<`, `>=`, `<=` — `rxGt`/`rxLt`/`rxGeq`/`rxLeq` in `R/d.R`) place their smoothed-derivative "bump" **off** the discontinuity, at `a - b ≈ 0.46` instead of at the step location `a - b = 0`. Because the forward pass evaluates the operator as a **hard** boolean, this makes the emitted derivative inconsistent with the value it differentiates: in a band next to the boundary the value is already saturated but the derivative is large and spurious. This corrupts any consumer that differentiates analytically through the sensitivity model (`.rxSens`) for models containing a parameter-dependent relational, `abs()`, `min()`, or `max()`.

The fix removes a shift term (`delta = atanh(2*tol - 1)`) from the four inequality-derivative rules, re-centering the bump on the discontinuity. `rxEq` is already centered and is unchanged.

## Background: how a relational is differentiated today

A relational `R(a,b) ∈ {0,1}` is discontinuous, so it is approximated for differentiation by a smooth sigmoid of `x = a - b`:

```
S(x) = 1/2 + 1/2 * tanh(k*x)          k = .rxD$..k = 10
```

and the derivative rule emits `dS/dx`, a **nascent delta** (a smooth stand-in for the Dirac that is the true derivative of a step):

```
dS/dx = (k/2) * sech^2(k*x) = k/2 - (k/2)*tanh(k*x)^2
```

For this to be a faithful derivative of a unit step, the bump must (i) sit **at the jump**, `x = 0`, and (ii) **integrate to 1** (the jump size).

## The problem (old mathematics)

The current rules use a **shifted** sigmoid,

```
S(x) = 1/2 + 1/2 * tanh(delta + k*x),   delta = atanh(2*tol - 1) ≈ -4.60512   (tol = 1e-4)
```

so, for example,

```
d(rxGt(a,b))/da  =  5 - 5*tanh(-4.60512 + 10*(a-b))^2
```

The shift makes the smoothed **value** read `S(0) ≈ tol ≈ 0` at equality (encoding strict `>` in the *smoothed* value). But it also **moves the derivative bump off the boundary**: `dS/dx` peaks where `delta + k*x = 0`, i.e.

```
x = a - b = -delta/k ≈ 0.46
```

The integral is still 1, but the mass is in the wrong place. And critically, the forward pass does **not** use `S` — it emits the hard boolean `(a > b)`. So the value jumps at `x = 0` while the derivative's mass sits at `x ≈ 0.46`. For any expression built on a relational, e.g. `max(a,b) = (a-b)*rxGt(a,b) + b`:

```
d(max)/da = rxGt(a,b) + (a-b) * d(rxGt)/da
```

the true value is `(a>b)`; the extra `(a-b)*bump` term should be negligible everywhere (it is `0` at the boundary because `(a-b)=0`, and `0` away from it because `bump→0`). With the **offset** bump it is instead large in a whole band `a-b ∈ ~(0.2, 0.7)`, where the model value is already constant. A consumer that differentiates through the operator then gets a derivative inconsistent with the function — measured as a ~4.7× error on a parameter-dependent branch, and up to ~400 −2LL units of estimate bias in a fit whose parameter happens to sit ≈0.46 past a threshold (details below).

## The fix (new mathematics)

Set `delta = 0` (drop the shift). The sigmoid becomes symmetric and the bump re-centers on the discontinuity:

```
S(x)   = 1/2 + 1/2 * tanh(k*x)                          S(0) = 1/2
dS/dx  = (k/2) * sech^2(k*x) = k/2 - (k/2)*tanh(k*x)^2   peak at x = 0, integral = 1
```

so the emitted derivatives become (k = 10):

```
                 OLD (offset)                              NEW (centered)
d(rxGt)/da   5 - 5*tanh(-4.60512 + 10*(a-b))^2        5 - 5*tanh(10*(a-b))^2
d(rxGt)/db  -5 + 5*tanh(-4.60512 + 10*(a-b))^2       -5 + 5*tanh(10*(a-b))^2
d(rxLt)/da  -5 + 5*tanh( 4.60512 + 10*(a-b))^2       -5 + 5*tanh(10*(a-b))^2
d(rxGeq)/da  5 - 5*tanh( 4.60512 + 10*(a-b))^2        5 - 5*tanh(10*(a-b))^2
d(rxLeq)/da -5 + 5*tanh(-4.60512 + 10*(a-b))^2       -5 + 5*tanh(10*(a-b))^2
```

`rxEq` is already centered (`d/da = -20*tanh(10*(a-b)) + 20*tanh(10*(a-b))^3`) and is untouched.

Two points worth noting:
- **No functional information is lost.** The smoothed *value* `S` is never emitted — the forward pass uses the hard boolean — so `S(0) = 1/2` vs `≈ tol` has no runtime effect. The strict/non-strict distinction (`>` vs `>=`) lives entirely in the hard value; the derivative (jump location and unit size) is identical for the pair, and both are now correctly centered.
- This also makes the **first derivative of `abs()`, `min()`, `max()` exact** (they are built from `rxGt`/`rxLt`): e.g. `d/da |a| = 2*(a>0) - 1 = sign(a)` with no near-zero smoothing error.

## Impact / who this affects

This rule feeds the variational sensitivity model (`.rxSens`). **Any consumer that differentiates analytically through a relational is affected**, and gets a derivative inconsistent with the hard forward value in a band ≈0.46 past the threshold:

- **External exact-AD sensitivity consumers** (e.g. aggregate/meta-analysis fitters that read `.rxSens` directly, including for unpaired structural parameters): gradient error through a parameter-dependent relational drops from a flat ~4.7× to a `sech²`-decaying term, negligible unless the parameter sits within ~0.2 of the threshold.
- **FOCEI, wherever its analytic gradient path is engaged** — an `eta` entering the branch (inner sensitivity), or the fast/analytic outer gradient. There the offset can drive a badly biased fit (measured up to ~400 −2LL units when the population sits in the displaced bump zone), and the centered rule recovers the correct fit.

Where the analytic path is *not* engaged (e.g. a purely finite-differenced outer gradient — the default `fast = FALSE` outer path for an unpaired parameter), the fit is unchanged: an unpaired-parameter branch is bit-identical before/after and equal to the branchless reference. In every case where the rule *is* exercised, the offset is biased and the centered rule is correct. A genuine discontinuity *straddled* by a random effect remains hard for the Laplace approximation either way — this change neither causes nor cures that.

## Validation

On data where every subject sits on one side of the branch, a branch model is **mathematically identical** to the equivalent branchless model, so a correct derivative must reproduce the branchless fit. Across `if/else`, `max`, `min`, `abs`, exercised through the analytic sensitivity path (an external aggregate-AD fitter, and FOCEI with an `eta` in the branch):

- **Centered** reproduces the branchless reference for all structures (Δ objective ≤ 0.7 −2LL).
- **Offset** biases exactly the structures whose threshold lies ≈0.46 past the population (`if/else`, `max`: +145…+400 −2LL, biased θ), and is harmless only where the population is far from its displaced bump (`min`, `abs`) or where the parameter is finite-differenced. This is the signature of a mislocated bump, not a structural issue.

A smoke test (plain PK, a parameter-dependent branch, and an `abs()` model) converges with finite standard errors under the centered rule.

## Tests

`tests/testthat/test-dsl.R` pinned the offset forms of the eight inequality derivatives; updated to the centered forms. `rxEq`'s assertions are unchanged. Full `test-dsl.R` passes (1010 expectations).

## Notes

The shift was a deliberate regularization (it encoded strict-vs-non-strict in the smoothed value and biased the derivative onto the "active" side of the branch). This change keeps the smoothing — a non-degenerate gradient at the branch, which the analytic paths rely on — but places it where the value actually transitions, so the derivative is consistent with the forward value for every consumer. Constants `..k` and `..tol` are unchanged (`..tol` no longer affects the inequality derivatives; it is retained in case other code reads it).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,17 @@
+# rxode2 (development version)
+
+## Bug fixes
+
+- The symbolic derivatives of the relational operators (`>`, `<`, `>=`, `<=`)
+  are now centered on the discontinuity `a == b`: the `atanh(2*tol - 1)` shift
+  that placed the smoothed nascent-delta bump at `a - b ~ +/-0.46` was removed.
+  Since the forward pass evaluates relationals as hard booleans, the shifted
+  bump gave sensitivity/exact-gradient consumers (e.g. FOCEI's analytic
+  gradient paths) a spurious derivative in a band next to the threshold; the
+  centered rule makes the derivative consistent with the forward value.  This
+  also makes the first derivatives of `abs()`, `min()`, and `max()` exact away
+  from the boundary (#1159).
+
 # rxode2 5.1.5
 
 ## New features
--- a/R/d.R
+++ b/R/d.R
@@ -423,7 +423,8 @@
 )
 
 ## Approx a==b by
-## (1-tanh(k*(a-b))^2)
+## (1-tanh(k*(a-b))^2)  -- a bump already centred at a==b, so its derivative
+## (the double lobe below) needs no shift.
 .rxD$rxEq <- list(
   function(a, b) {
     .ab <- paste0("(", a, "-", b, ")")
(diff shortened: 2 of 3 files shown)


mattfidler:
Reviewed this locally three ways -- my own review (Claude), plus independent reviews by GitHub Copilot (GPT-family) and Antigravity (Gemini-family) over the checked-out branch. All three reviewers independently verified the mathematics:

- All eight re-centered derivative strings have the correct signs (`d/da` positive for `>`/`>=`, negative for `<`/`<=`, mirrored for `d/db`); the bump peaks at `a==b` and integrates to 1.
- Numerically, the old rule's spurious `(a-b)*bump` term peaked at ~2.3 at 0.46 past the boundary; the centered rule reduces this to the inherent smoothing floor (~0.22 at the boundary itself).
- Nothing else in rxode2 (or in nlmixr2est / nlmixr2 / babelmixr2 / nonmem2rx / monolix2rx) pins the old shifted strings.
- `abs()`/`min()`/`max()`/SELU build symbolically on `rxGt`/`rxLt` and inherit the fix; verified end-to-end through `rxS()`. A nice invariant falls out: the `max` and `min` bump terms cancel exactly, so `d/da (max+min) = 1` as it must.

No correctness bugs found. I pushed a follow-up commit (e3c8bc0) addressing the review findings:

1. **Removed `.rxD$..tol`** (flagged by both external reviewers). After this fix nothing reads it -- I grepped rxode2 and the reverse dependencies -- so keeping it would make `.rxD$..tol <- <x>` a silent no-op for anyone trying to tune the smoothing. Dead-but-plausible-looking knobs are worse than absent ones.
2. **Pinned the composite `max`/`min`/`abs` derivatives** in `test-dsl.R` (flagged by both external reviewers): the PR only pinned the primitive `rxGt`/`rxLt` strings, so a regression in the composite expansion (the actual user-facing behavior this PR fixes) would have passed silently. The new test goes through `rxS()`/symengine, the same path `.rxSens` uses.
3. **Added the missing `NEWS.md` entry** under a development-version heading -- this changes gradient behavior for any model with a parameter-dependent branch, so it needs user-facing release notes.
4. **Comment style**: condensed the new comment blocks to the repo's terse style and switched `centred`/`centres` to American English per repo convention.

Validation after the changes, all in a clean worktree: `test-dsl.R` passes (1013 expectations, was 1010), and the sensitivity/jacobian/adjoint suites -- the code paths that actually consume these derivative rules -- pass (3015 expectations, 0 failures).

Nice catch on the displaced bump -- the analysis in the PR description is exactly right.

PR #1160, opened 2026-07-24
Fix the sign of the Yeo-Johnson derivatives at lambda == 2

hiddevandebeek:
## Summary

The `lambda == 2` special cases in the Yeo-Johnson derivative code return the
derivatives with the wrong sign on the negative branch.

For `x < 0` the Yeo-Johnson transform at `lambda == 2` is `yj(x) = -log(1 - x)`,
so

* `yj'(x)  = 1/(1 - x)`
* `yj''(x) = 1/(1 - x)^2`

both strictly positive. `inst/include/rxode2.h` returned `-1/(1 - x)` and
`-1/((1 - x)*(1 - x))`.

## Why the special cases are wrong, not the general formulas

Two checks that do not depend on redoing the derivation:

**1. They contradict their own general formula in the limit.** The lines
immediately following each special case handle general `lambda` and are correct:

* `pow(1 - x, 1 - lambda)` at `lambda = 2` is `(1 - x)^-1 = +1/(1 - x)`
* `-(1 - lambda)*pow(1 - x, -lambda)` at `lambda = 2` is `+1/(1 - x)^2`

Both have the opposite sign to the special case sitting directly above them, so
the derivative was discontinuous in `lambda` at exactly 2. On 5.1.4, at `x = -3`:

| lambda | `rxTBSd()` |
|---|---|
| 1.999999 | +0.25000035 |
| **2.0** | **-0.25** |
| 2.000001 | +0.24999965 |

**2. Yeo-Johnson is monotone increasing**, so its first derivative must be
positive for every `x` and every `lambda`. A negative value is impossible
regardless of any derivation.

## Reproduction

```r
m <- function() {
  model({
    v  <- rxTBS(xv, lam, 1, 1, 0)   # yj = 1 -> yeoJohnson
    d1 <- rxTBSd(xv, lam, 1, 1, 0)
    d2 <- rxTBSd2(xv, lam, 1, 1, 0)
  })
}
et <- et(1:3); et$xv <- c(-3, -1, -0.5); et$lam <- 2
rxSolve(m, et)[, c("v", "d1", "d2")]
```

Before:

```
          v         d1          d2
 -1.3862944 -0.2500000 -0.06250000
 -0.6931472 -0.5000000 -0.25000000
 -0.4054651 -0.6666667 -0.44444444
```

The values are right; both derivatives are negated. The closed form gives
`+0.25, +0.50, +0.667` and `+0.0625, +0.25, +0.4444`, which is also what a
central finite difference of `v` returns.

## Changes

Two characters, in `inst/include/rxode2.h`:

* `_powerDD()` (Yeo-Johnson first derivative, `x < 0`, `lambda == 2`):
  `-1/(1.0 - x)` becomes `1/(1.0 - x)`
* `_powerDDD()` (second derivative, same branch):
  `-1/((1.0 - x)*(1.0 - x))` becomes `1/((1.0 - x)*(1.0 - x))`

The composite `logit`-then-Yeo-Johnson and `probit`-then-Yeo-Johnson transforms
(`yj` cases 5 and 7) call this code recursively rather than repeating the
special case, so they are corrected by the same change. The forward transform
and the inverse were already correct, as are the `x >= 0` special cases at
`lambda == 0`, which agree with their general formula's limit.

## Scope and reachability

This is reachable only when `lambda` is **exactly** `2.0` -- every other value
takes the correct general branch. I did not find any code path that bounds
`lambda` at 2, so an optimiser will not land there by itself; the realistic
route is a user fixing `yeoJohnson(2)`. Low impact, but a clear correctness bug
with a two-character fix.

## Verification

Built from source and checked against the patched build:

* At `lambda = 2` on the negative branch, `rxTBSd()` and `rxTBSd2()` now match
  the closed forms `1/(1 - x)` and `1/(1 - x)^2` with an error of exactly 0.
* Both derivatives match central finite differences of `rxTBS()` at
  `lambda` in {0, 0.5, 1.5, 1.999999, 2, 2.000001} over `x` in
  {-3, -1, -0.5, 0.5, 2}, spanning both branches.
* The discontinuity in `lambda` is gone: the gap between `lambda = 2` and
  `lambda = 2 +/- 1e-6` falls from 1.33 to 3.5e-7 for the first derivative and
  from 0.89 to 2.6e-7 for the second.
* `rxTBSd()` is positive everywhere tested, as a monotone transform requires.
* `tests/testthat/test-dsl.R`: 1008 expectations, 0 failures.

## Tests

No existing test pinned these values, so nothing had to be updated -- but that
is also why the sign error survived. Adds
`tests/testthat/test-yeojohnson-derivative.R` (17 expectations) covering the
closed form at `lambda == 2`, continuity in `lambda` across the special case,
and finite differences on both branches for several `lambda`.

Built and tested on Windows, R 4.5.3.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -43,6 +43,14 @@
 
 ## Bug fixes
 
+- The first and second derivatives of the Yeo-Johnson transform (`rxTBSd()` and
+  `rxTBSd2()`) had the wrong sign for negative values when `lambda` was exactly
+  `2`.  There `yj(x) = -log(1 - x)`, so the derivatives are `1/(1 - x)` and
+  `1/(1 - x)^2`, both positive; the special case returned them negated, which
+  also contradicted the general formula's limit and made the derivative
+  discontinuous in `lambda` at `2`.  Since Yeo-Johnson is monotone increasing,
+  its first derivative must be positive everywhere.
+
 - The `parsed_md5` of a model no longer depends on how many models were built
   before it in the session.  `linCmtSens` was folded into the hash but only
   assigned *after* the model was parsed, so the first build of a session hashed
--- a/NEWS.md
+++ b/NEWS.md
@@ -70,6 +70,12 @@
     missing the `probit + yeoJohnson` case (returned `NA`).  The log-Jacobian
     itself (`powerL`) clamped the wrong term in its `logit` guard, giving an
     unprotected `log(0)` at the upper bound.
+  - For `boxCox`/`lnorm`, `rxTBSd()` and `rxTBSd2()` returned the clamp
+    constant `sqrt(.Machine$double.eps)` itself for `x` at or below the clamp
+    instead of clamping `x` and evaluating the derivative formula, making the
+    derivatives discontinuous (and ~15 orders of magnitude too small) at the
+    boundary.  The clamp now feeds the usual formula, matching how every other
+    transform in the family handles the guard.
 
 - The `parsed_md5` of a model no longer depends on how many models were built
   before it in the session.  `linCmtSens` was folded into the hash but only
--- a/NEWS.md
+++ b/NEWS.md
@@ -66,10 +66,11 @@
   - The `lambda` gradient of the transform log-Jacobian (`powerDL`, used by
     estimation routines) was wrong on the negative Yeo-Johnson branch
     (`-log1p(x)` instead of `-log1p(-x)`, `NaN` for `x < -1`), returned a
-    spurious `0` at exactly `lambda == 1` for `boxCox`/`yeoJohnson`, and was
-    missing the `probit + yeoJohnson` case (returned `NA`).  The log-Jacobian
-    itself (`powerL`) clamped the wrong term in its `logit` guard, giving an
-    unprotected `log(0)` at the upper bound.
+    spurious `0` at exactly `lambda == 1` for `boxCox`/`yeoJohnson`, was
+    missing the `probit + yeoJohnson` case (returned `NA`), and returned a
+    spurious `log(x)` (instead of `0`) for the lambda-free `lnorm` transform.
+    The log-Jacobian itself (`powerL`) clamped the wrong term in its `logit`
+    guard, giving an unprotected `log(0)` at the upper bound.
   - For `boxCox`/`lnorm`, `rxTBSd()` and `rxTBSd2()` returned the clamp
     constant `sqrt(.Machine$double.eps)` itself for `x` at or below the clamp
     instead of clamping `x` and evaluating the derivative formula, making the
--- a/NEWS.md
+++ b/NEWS.md
@@ -51,6 +51,26 @@
   discontinuous in `lambda` at `2`.  Since Yeo-Johnson is monotone increasing,
   its first derivative must be positive everywhere.
 
+- Review of the fix above found further errors in the same transform family
+  (all pre-existing, none introduced by that fix):
+  - `rxTBSd2()` returned a wrong second derivative for the `logit` transform
+    (an algebra error in the closed form) and for the composed
+    `logit + yeoJohnson` / `probit + yeoJohnson` transforms, where the chain
+    rule used the first Yeo-Johnson derivative in place of the second and
+    dropped the inner-transform curvature term.
+  - `rxTBSi()` did not invert the composed `logit + yeoJohnson` /
+    `probit + yeoJohnson` transforms: it applied the forward Yeo-Johnson
+    transform (or skipped it entirely) instead of the inverse, so
+    `rxTBSi(rxTBS(x))` did not return `x` for `lambda != 1`.  This affected
+    simulation back-transforms of those error models.
+  - The `lambda` gradient of the transform log-Jacobian (`powerDL`, used by
+    estimation routines) was wrong on the negative Yeo-Johnson branch
+    (`-log1p(x)` instead of `-log1p(-x)`, `NaN` for `x < -1`), returned a
+    spurious `0` at exactly `lambda == 1` for `boxCox`/`yeoJohnson`, and was
+    missing the `probit + yeoJohnson` case (returned `NA`).  The log-Jacobian
+    itself (`powerL`) clamped the wrong term in its `logit` guard, giving an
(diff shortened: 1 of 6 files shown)


mattfidler:
Thanks @hiddevandebeek -- this fix is correct and the analysis in the PR description holds up.

This branch was put through three independent reviews (Claude, GitHub Copilot/GPT, and Antigravity/Gemini). All three confirmed the change:

* For `x < 0`, `yj(x) = -((1-x)^(2-lambda) - 1)/(2-lambda)`; at `lambda == 2` this is `-log(1-x)`, so `yj' = 1/(1-x)` and `yj'' = 1/(1-x)^2`, both positive -- matching the general-formula limits and restoring continuity in `lambda`.
* The composite `logit`/`probit + yeoJohnson` first derivatives (yj codes 5/7) recurse into the fixed case, as the description says.
* The inverse `_powerDi` at `lambda == 2` (`-expm1(-x)`) was already correct.
* New test file passes (17 expectations), is ASCII-only, and properly wrapped in `rxTest({})`.

## Additional bugs found in the same family (fixed in 52b7fd52f)

The reviews were pointed at the whole `_power*` family, and they converged on several *pre-existing* errors (none introduced here). Since they live in the same functions and the new test scaffolding covers them, I pushed a fix commit to this branch:

1. **`_powerDDD` case 4 (logit second derivative)**: the code implemented `hl^2/(hl^4*t1^2)` where its own comment had `hl^2/(xl^4*t1^2)`; at `x = 0.25` on `(0,1)` it returned `-42.56` instead of `logit''(0.25) = -14.22`. Rewritten as the equivalent `1/(high-x)^2 - 1/(x-low)^2`.
2. **`_powerDDD` cases 5/7 (composite second derivative)**: the chain rule used the *first* Yeo-Johnson derivative where the second belongs and dropped the `f'(g)*g''` term. At `lambda = 1` case 5 returned `logit'(x)^2` (`+28.4` at `x = 0.25`) instead of `logit''(x)` (`-14.2`). Now `f''(g)*g'^2 + f'(g)*g''`.
3. **`_powerDi` cases 5/7 (composite inverse, `rxTBSi`)**: applied the *forward* Yeo-Johnson transform where the inverse belongs, and case 7 discarded even that and returned `pnorm` of the raw input -- so `rxTBSi(rxTBS(x)) != x` for `lambda != 1`. This affected simulation back-transforms of `logit`/`probit + yeoJohnson` error models. Both now reuse the correct case-1 inverse.
4. **`_powerDL` (the `d(logLik)/dlambda` exported as `powerDL`)**: on the negative branch returned `-log1p(x)` instead of `-log1p(-x)` (NaN for `x < -1`); had a spurious `lambda == 1 -> 0` early return for boxCox/yeoJohnson (a `lambda` started at exactly 1 got a zero gradient); and was missing case 7 entirely (probit+YJ returned `NA`).
5. **`_powerL` case 4**: the clamp guard tested `xl` instead of the `hl2` it clamps, leaving `log(0)` unprotected at `x == high`.

Verification: R-level FD/round-trip tests added to `test-yeojohnson-derivative.R` for the model-reachable functions (now 57 expectations, all passing along with `test-dsl.R` and `test-logit.R`, 3245 total); `powerL`/`powerDL` are not reachable from the model language, so they were verified with a standalone compile of the header against central differences and the `powerL == log(powerDD)` identity.

## Flagged but intentionally not changed

`_powerDD` (cases 0/3) and `_powerDDD` (case 0) contain `if (x <= _eps) return x0 = _eps;` -- returning `1e-6` where the surrounding code pattern suggests clamping `x` and evaluating the formula (which would return ~`1e6`/`-1e12` at the boundary). It looks like a typo, but "fixing" it changes boundary behavior by ~12 orders of magnitude and may be accidentally load-bearing for estimation stability, so it is left as a separate decision.

mattfidler:
Follow-up on the deferred item from the review above: the `if (x <= _eps) return x0 = _eps;` clamps have now been addressed, plus two more rounds of independent review (Copilot/GPT and Antigravity/Gemini) over the growing branch.

## Commit `d6de3d295` -- clamp typos

`_powerDD` (cases 0/3, boxCox/lnorm) and `_powerDDD` (case 0) returned the clamp constant `sqrt(DBL_EPSILON)` (~1.5e-8) itself for `x` at or below the clamp, instead of clamping `x` and evaluating the derivative formula (~6.7e7 for `1/x` at the boundary) the way every other guard in the family does. The clamp itself is kept -- it exists for stability and these paths are rarely hit -- but it now feeds the formula, making the derivatives continuous at the boundary. Behavior checks pin the clamped values and continuity. **The full test suite passes unchanged (0 failures, 79143 passes), so nothing relied on the old sentinel return.**

## Commit `3da7b7e4b` -- re-review findings

Both re-reviewers independently converged on the same remaining bug: `_powerDL` case 3 (lnorm) returned `log(x)`, but the lnorm log-Jacobian `-log(x)` has no `lambda` dependence at all, so the gradient must be `0` like the other lambda-free transforms. Fixed.

They also both flagged that `powerL`/`powerDL` had zero test coverage -- they are only shared via `R_RegisterCCallable()` (nlmixr2est) and were unreachable from R, which is exactly why every bug in them survived. This commit adds an internal `.Call` hook (`_rxode2_powerLDL`, R side `rxode2:::.rxTransformL`) and tests pinning the log-Jacobian closed forms, the lambda gradient (including at exactly `lambda == 1` and on the negative Yeo-Johnson branch), zero gradients for the lambda-free transforms, the composed logit/probit chains, and agreement with central differences of `powerL` over `lambda`. The finite-difference derivative test loop also now covers plain probit (`yj = 6`).

## Commit `8b869eb71` -- final round

A last review round over the complete branch: Antigravity found nothing (it verified the wrapper's protection/arity/validation, the case-3 math, and that the downstream `powerDL` CCallable signature is unchanged); Copilot's single finding was that the new wrapper skipped the `high > low` bounds check its sibling `_rxode2_powerD` performs -- mirrored and pinned with a test.

State of the branch: `test-yeojohnson-derivative.R` is at 94 expectations, all green; full suite 0 failures / 79143 passes; three independent reviewers have no outstanding findings.

PR #1162, opened 2026-07-24
Load RcppParallel before rxode2's DLL to fix Windows load failure (RcppParallel >= 6.0.0)

billdenney:
## Problem

On Windows, `library(rxode2)` (and installation, at the load-test step) can fail with:

```
unable to load shared object '.../rxode2/libs/x64/rxode2.dll':
  LoadLibrary failure:  The specified module could not be found.
```

`rxode2.dll` dynamically imports `RcppParallel.dll` (it links with `-lRcppParallel`). Windows can only resolve that import by base name if `RcppParallel.dll` is already loaded in the process. Because `RcppParallel` was listed only under `LinkingTo` (not `Imports`, and with no `importFrom`), R never loaded the RcppParallel namespace/DLL before loading rxode2's DLL, and its `libs/x64` directory is not on the DLL search path → `ERROR_MOD_NOT_FOUND`.

This became visible with **RcppParallel 6.0.0**, which statically links TBB into `RcppParallel.dll` and no longer ships the `tbb.dll` stub. Previously, the stale `-L.../RcppParallel/lib/x64 -ltbb -ltbbmalloc` flags (since stripped by #1161) happened to drag the library in. #1161 fixed the *link*; this fixes the *load*.

## Fix

Make RcppParallel a runtime dependency so its namespace (and DLL) loads ahead of rxode2's:

- Add `RcppParallel` to `Imports` in `DESCRIPTION` (kept in `LinkingTo`).
- Add `importFrom(RcppParallel, RcppParallelLibs)` (via roxygen in `R/rxode2.R`).

This is the standard RcppParallel-on-Windows pattern.

## Verification (R 4.6.1, RcppParallel 6.0.0, StanHeaders 2.32.10, Rtools45)

- `R CMD INSTALL` completes including the "can this package be loaded" test (previously failed there).
- A clean session with no manual preload: `library(rxode2)` loads, and `"RcppParallel" %in% loadedNamespaces()` is `TRUE`.
- A one-compartment model solves correctly.

Before this change, the same clean-session `library(rxode2)` failed with the `LoadLibrary` error above.


mattfidler:
This is already fixed in main.

mattfidler:
is different.

PR #1170, opened 2026-07-30
fix: make ui$modelName a single character string

### LINKED ISSUE #1019: ui$modelName should always be a scalar character string
While adding tons of specific models to nlmixr2lib, I came across an issue where the modelName can be a character vector. I think that the intent is that it's always a single character string. If that thought is correct, can we fix this so that it uses a way to always return a single character string?

```
library(nlmixr2)
library(nlmixr2lib)
mod <- rxode2(readModelDb("PK_1cmt"))
mod$modelName
# [1] "readModelDb" "PK_1cmt"    
```

billdenney:
as.character() of the substituted model expression returns one element per part of a call, so rxode2(readModelDb("PK_1cmt")) stored c("readModelDb", "PK_1cmt") and an anonymous model function stored a four-element vector that included the deparsed body.

.rxModelNameFromExpr() now collapses the expression to one name: a symbol keeps its name, a call is deparsed to a single string, and an anonymous function (with or without the `(` needed to call it) has no name and gives NULL.  The empty symbol from a missing argument stays in that frame -- handing it to another closure raises "argument is missing" -- and resolves to NULL.

rxUiGet.modelName normalizes on read as well, so vectors written by released reverse dependencies (nlmixr2est's nlmixr2()/nlmixr2.rxUi() use the same as.character(substitute()) pattern) and values in models saved by earlier versions also come back as a single string.

Fixes #1019.


billdenney:
@mattfidler, This is ready for review.

mattfidler:
This kind of breaks the assumptions from `nlmixr2est` I think.  I would also like this to eventually work with things like `:=` in `nlmixr2save`, but shouldn't really be owned here.

It probably should be made into a s3 method based on the function called (with a default)

billdenney:
For my knowledge, what is the assumption within `nlmixr2est` that this breaks? I thought that the model name was mostly convenience/viewed rather than being used anywhere.

mattfidler:
No, it is also used in nlmixr2est and derives the assignment

```
fit <- nlmixr2(model)
```

The underlying model name `model` will be ported.  It is also used in babelmixr2, monolix2rx etc.  There are many consumers and I don't think this may break them.

mattfidler:
In particular in babelmixr2 this drives the default export for NONMEM and Monolix

mattfidler:
I was going to fix this in the next relase, but I need to submit to CRAN today (before the CRAN shutdown), so my fixes have to be minimal and I don't know how this affects downstream packages yet 🫤

mattfidler:
@billdenney 

This may work a bit better for your use case.  Let me know what you think.

billdenney:
As long as it's one character string, it works for me.

mattfidler:
You could make it so you get the method through a s3 extension, which would mean you simply use `$modelName` and it will give you the right value.

I guess you are simply trading one item for another.

PR #1174, opened 2026-08-04
Fix Windows build with new RcppParallel

jeroen:
The package no longer builds on Windows because RcppParallel 6.x dropped the TBB library on Windows.

See https://cran.r-project.org/web/checks/check_results_rxode2.html

As a result packages that depend on rxode2  also fail to load:

https://cran.r-project.org/web/checks/check_results_nlmixr2.html

This PR is the solution is that Claude came up with, and it works but maybe there is something better?

cc @kevinushey

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: rxode2
-Version: 5.1.5
+Version: 5.1.6
 Title: Facilities for Simulating from ODE-Based Models
 Authors@R: c(
    person("Matthew L.","Fidler", role = c("aut", "cre"), email = "matthew.fidler@gmail.com", comment=c(ORCID="0000-0001-8538-6691")),
@@ -87,7 +87,7 @@ Suggests:
     tidyr,
     usethis,
     withr,
-    xgxr,
+    xgxr (>= 1.1.6),
     pillar,
     tibble,
     units (>= 0.6-0),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,45 @@
+# rxode2 5.1.6
+
+## Bug fixes
+
+### Dependencies
+
+- The suggested `xgxr` is now required to be `>= 1.1.6`.  Its
+  `xgx_scale_x_log10()`/`xgx_scale_y_log10()` return the `ggplot2` scale
+  itself from that version on, rather than a length-one list wrapping it.
+
+### Installation / linking
+
+- Fixed the Windows build against `RcppParallel` 6.2.0, which no longer ships
+  the TBB library there (6.0.0 still built).  `configure` already dropped the
+  `-ltbb`/`-ltbbmalloc`
+  link flags when they are unavailable, but still compiled with `-DSTAN_THREADS`
+  and `-DRCPP_PARALLEL_USE_TBB=1`, which pulls `stan::math`'s `ad_tape_observer`
+  (a `tbb::task_scheduler_observer`) into the objects and left undefined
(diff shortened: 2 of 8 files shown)


mattfidler:
I ported this right as RcppParallel was being put onto CRAN.  It did work with some versions.  

Is RcppParallel now stable?  I have already had multiple submissions to CRAN (3 submissions in 2 weeks) due to this (and I suppose other things).  I am afraid they will start banning me because I'm spamming them.

mattfidler:
I will also have to check if it kills the parallel nlmixr2est before continuing I guess. 

Thanks for the heads up;  I guess CRAN really wants this to work for packages.  Although I haven't been told I would be banned, I will fix it again.

mattfidler:
This also breaks my tests, so I guess more work will need to be done.

mattfidler:
It was working before `RcppParallel` 6.2.0 which was released 5 days ago as my previous successful pull requests (and the release date of `RcppParallel` shows.

jeroen:
I don't think they will ban you for fixing things on short notice :) It has just been a bumpy update.

mattfidler:
Ok.  Thanks for the heads up.

kevinushey:
Sorry for the trouble here @mattfidler -- this churn is my fault, and I apologize for the extra CRAN submissions it's cost you. Thanks @jeroen for the fix and the ping.

What happened: RcppParallel changed its Windows TBB arrangement twice in about a week. 6.0.0 stopped shipping a TBB library there (it linked the static Rtools TBB into RcppParallel.dll instead), and rxode2 correctly adapted by stripping `-ltbb -ltbbmalloc`. Then 6.2.0 reverted course -- the Rtools TBB turned out to be a dead end -- and we now build and ship the bundled oneTBB (`tbb.dll` + `tbbmalloc.dll`), same as on every other platform. That left the 6.0.0-era workaround stripping flags that now point at a real library, hence the undefined `tbb::detail::r1::observe` at link time.

So one note on the NEWS entry here: 6.2.0 restored the TBB library on Windows rather than dropping it. If losing `STAN_THREADS` on Windows ends up hurting parallel nlmixr2est, linking against the flags from `RcppParallelLibs()` should now just work -- happy to send a PR for that.

> Is RcppParallel now stable?

Yes -- 6.2.0 is the arrangement I should have landed on in 6.0.0 directly, and I have no further linking changes planned. My revdep checks run on Linux, which is how a Windows-only breakage sailed through; before any future release that touches linking, I'll build the nlmixr2 stack on Windows first and give you a heads up.

And to echo Jeroen -- CRAN isn't going to ban you for fixing breakage another package caused. Happy to write to them myself if that would help.

mattfidler:
I don't use the `RcppParallel` directly now, I thread on OpenMP, so this works fine.  

While I was thinking of changing to RcppParallel (because of the mac support of openmp), I was reading that it isn't quite as straight forward as a lift and shift, so I haven't done it yet.

I was just thinking the "submitting too many times"/year 

I will submit soon though.  Need to verify that everything I need for the next submission is live.

jeroen:
Just note that cran is closed for submissions starting tomorrow, see
https://cran.r-project.org/

mattfidler:
Oh my!  this may be delayed (I hope not).

kevinushey:
Follow-up PR as promised: #1178 restores TBB/STAN_THREADS on Windows with RcppParallel >= 6.2.0 (and corrects the release notes here). No rush -- it's meant for the next regular release, not this CRAN cycle.

mattfidler:
Submitted earlier.

Thanks for the update

PR #1178, opened 2026-08-04
keep tbb and stan_threads on windows with rcppparallel >= 6.2.0

kevinushey:
Follow-up to #1174 (cc @mattfidler @jeroen) -- and to be clear up front: **no rush on this one**. 5.1.6 is correct to ship as-is; this is for whenever the next regular release happens.

## What this changes

RcppParallel 6.2.0 ships `tbb.dll`/`tbbmalloc.dll` on Windows again (in `RcppParallel/lib/x64/`), so the `-ltbb -ltbbmalloc` flags emitted by `StanHeaders:::LdFlags()` and `RcppParallel:::RcppParallelLibs()` point at a real, linkable library there. The 5.1.6 NEWS had this backwards -- it was 6.0.0--6.1.1 that shipped no TBB library on Windows (those versions linked the static Rtools TBB into `RcppParallel.dll`); 6.2.0 restored it. Sorry again for the churn that made this hard to track.

This PR makes `workaround.R` decide whether to strip the TBB flags by looking for the TBB library on disk (`tbb*.dll` / `libtbb*.dll.a` / `libtbb*.a` under `RcppParallel/lib`) instead of by the shape of the `-L` flags, which can't distinguish 6.0.0's stale flags from 6.2.0's correct ones. Net effect:

* RcppParallel >= 6.2.0 on Windows: flags kept, so `STAN_THREADS` and the TBB link are restored (relevant if the TBB-less build turns out to matter for parallel nlmixr2est).
* RcppParallel 6.0.0--6.1.1: no TBB found, so the TBB-less arrangement from #1174 is used, exactly as today.
* The compat-header machinery from #1174 is untouched; it self-disarms once `STAN_THREADS` is defined again.

Also corrects the comments in `rxode2_sundials_stan_compat.h` / `linCmt.cpp` and adds a NEWS entry noting the correction.

## Testing

* Full R-CMD-check matrix green on my fork, including `windows-latest (release)`: https://github.com/kevinushey/rxode2/actions/runs/30920875113
* Verified the Windows job took the TBB path (not just that it passed): a diagnostic run on the same runner image confirmed the RcppParallel 6.2.0 binary ships `lib/x64/tbb.dll`, the new gate evaluates to keep-the-flags, and the build compiled with `-DSTAN_THREADS -DRCPP_PARALLEL_USE_TBB=1` and linked `-ltbb -ltbbmalloc`.
* Runtime DLL resolution is safe because rxode2's namespace imports RcppParallel, which loads `tbb.dll` into the process before `rxode2.dll` is loaded.

Happy to adjust anything here -- and if you'd rather keep Windows TBB-less for now, the NEWS/comment corrections could land on their own.


mattfidler:
Thanks @kevinushey

PR #1181, opened 2026-08-05
fix: rxCompile() must generate code for the model it is given

mattfidler:
## Problem

`_rxode2_codegen` emits C from the parser's **global** `.rxModelVarsLast`, not from the model variables passed to `rxCompile()`. The guard before it only asked whether *some* model was loaded:

```r
if (.Call(`_rxode2_codeLoaded`) == 0L) .rxModelVarsCharacter(setNames(.mv$model, NULL))
```

So whenever a *different* model happens to be the parsed one, the C written out is that other model's, while the artifact keeps this model's md5-derived name — and the `modVars` read back from the DLL describe the wrong model entirely.

`rxode2()` never hits this: it parses, then compiles immediately. A **re-compile** does. An existing model object whose `.so` is gone reaches `rxCompile()` via `rxDynLoad()` → `$compile()` long after some other model was parsed — which is exactly what happens to a saved fit restored in a new session, since its DLL lived in the original session's `tempdir()`.

Measured on a restored SAEM fit (nlmixr2's `broom` article, which loads a cached fit):

```
requested mv params:  tcl, tv
.rxModelVarsLast:     tcl, eta.cl, tv, eta.v     <- what got compiled
```

and every solve after that failed with

```
The following parameter(s) are required for solving: eta.v, eta.cl
```

## Fix

Re-parse unless the loaded model is provably this one (`parsed_md5` comparison). Only the model *syntax* (`normModel`) is parsed — the sibling `indLin` slot holds generated matrix-exponential C and is restored from `.indLinInfo`/`.rxMECode` a few lines below anyway.

In the ordinary `rxode2()` path the md5s match, so nothing re-parses and behavior is unchanged.

## Tests

New `tests/testthat/test-recompile-model.R`, covering both the direct `rxCompile()` call and the `rxLoad()`-after-the-`.so`-is-gone path. Verified the tests fail against the old guard (emulated by neutering the re-parse) — model A came back reporting model B's parameters — and pass with the fix.

Downstream check: nlmixr2's `broom` article, which failed to build on `setOfv(fit.s, "gauss3_1.6")`, now renders, and the full nlmixr2 article precompute (10 cached articles) renders with 0 failures.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,19 @@
 # rxode2 5.1.7
 
+## Bug fixes
+
+- `rxCompile()` now re-parses the model it is handed whenever the parser's
+  current model is a different one.  Code generation reads the parser's global
+  model state, and the old guard only checked whether *some* model was loaded,
+  so a re-compile requested while an unrelated model was parsed wrote that other
+  model's C under this model's name and handed back its model variables.
+  Building a model with `rxode2()` never hit this (it parses, then compiles
+  immediately), but re-loading one whose `.so` is gone did -- as when a saved
+  fit is restored in a new session, since its DLL lived in the original
+  session's `tempdir()`.  Such a fit came back solving a different model, e.g. a
+  restored SAEM fit failing with "The following parameter(s) are required for
+  solving: eta.v, eta.cl".
+
 # rxode2 5.1.6
 
 ## New features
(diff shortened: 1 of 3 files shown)


mattfidler:
Full test suite run against this branch (`pkgload::load_all()`, `NOT_CRAN=true`, 234 test files including the new `test-recompile-model.R`): **0 failures**.

PR #1194, opened 2026-08-06
fix(dde): optExpression and past() durations for delay differential equations

### LINKED ISSUE #1192: rxOptExpr(): 'unsupported lhs in optimize expression' for a DDE whose past()/delay() duration is an expression
`rxOptExpr()` cannot optimize a DDE whose `past()`/`delay()` duration is an
expression: it dies with `unsupported lhs in optimize expression`, so
`optExpression = TRUE` is unusable for that class of model.

This is why the DDE vignette in `nlmixr2` has to pass `optExpression = FALSE`
to every DDE fit.

## Minimal reproducible example (rxode2 only)

```r
library(rxode2)   # 5.1.7

mod <- "
lT=2.549445
a=1
b=0.5
k3=5
k1=0.535
k2=0.352
G(0)=a
d/dt(G)=k3-(k1/k2)*(1-exp(-k2*t))*G
past(G,exp(lT))=a*exp(b*t)
gdel=delay(G,exp(lT))
"

## the model itself is fine
s <- rxSolve(rxode2(mod), et(seq(0, 12, by = 1)), method = "dop853", dense = TRUE)
head(s$gdel, 3)
#> [1] 0.001661559 0.002739448 0.004516586

## optimizing it fails
rxOptExpr(mod)
#> -> finding duplicate expressions in model...
#> exp(lT)
#> Error in ..rxOptLhs(x[[3]]) : unsupported lhs in optimize expression
#> Error: stopped optimizing duplicate expressions
```

## End-to-end: the same thing kills an `nlmixr2()` fit

A DDE written with the delay inline (rather than through an intermediate
variable) fails the moment `optExpression` is left at its default:

```r
mod6b <- function() {
  ini({ tk4 <- log(0.15); tk5 <- log(0.10); tT <- log(11); tb <- log(0.40)
        prop.pk <- 0.382; add.tas <- 1.49; add.aks <- 0.420 })
  model({
    k <- 0.184; V <- 0.039; k1 <- 0.535; k2 <- 0.352; k3 <- 5
    k4 <- exp(tk4); k5 <- exp(tk5); Emax <- 1; EC50 <- 44.3
    I0 <- 2.42; a <- 1; b <- exp(tb)
    d/dt(Ac) <- -k * Ac
    c   <- Ac / V
    eff <- (Emax * c) / (EC50 + c)
    G(0) <- a
    d/dt(G) <- k3 - eff * G - (k1 / k2) * (1 - exp(-k2 * t)) * G
    past(G, exp(tT)) <- a * exp(b * t)
    I(0) <- I0
    d/dt(I) <- k4 * G - k4 * delay(G, exp(tT))
    D(0) <- 0
    d/dt(D) <- k4 * delay(G, exp(tT)) - k5 * D
    TAS <- I + D
    AKS <- D
    c   ~ prop(prop.pk); TAS ~ add(add.tas); AKS ~ add(add.aks)
  })
}

nlmixr2(mod6b, dat6, est = "focei",
        control = foceiControl(optExpression = TRUE))
#> Error : stopped optimizing duplicate expressions

nlmixr2(mod6b, dat6, est = "focei",
        control = foceiControl(optExpression = FALSE))
#> runs
```

This also means the estimation path is exposed generally: whenever symengine
inlines the delay duration into the `past()` line (e.g. `past(G,
exp(THETA[3] + ETA[1]))`), the assembled model can no longer be optimized.

## Which durations are accepted

Same model, varying only the `past()`/`delay()` duration:

| duration | `rxOptExpr()` |
|---|---|
| `lT` (bare name) | ok |
| `12.8` (number) | ok |
| `(lT)` | ok |
| `lT/2` | ok |
| `exp(lT)` | **fails** |
| `lT*2` | **fails** |
| `exp(THETA[1]+ETA[1])` | **fails** |

## Cause

`..rxOptLhs()` (`R/rxOptExpr.R:169`) recurses into the `past()` duration with
`..rxOptLhs()` itself:

```r
} else if (identical(x[[1]], quote(`past`))) {
  return(paste0("past(", ..rxOptLhs(x[[2]]), ",", ..rxOptLhs(x[[3]]), ")"))
```

but `..rxOptLhs()` only understands atomics, names, and a fixed set of
left-hand-side calls (`dt`, `f`, `F`, `rate`, `alag`, `lag`, `dur`, `past`,
`dy`, `df`, `/`, `(`). Anything else -- which is any real duration expression --
falls through to `R/rxOptExpr.R:178`:

```r
  print(x)
  stop("unsupported lhs in optimize expression")
```

`/` and `(` only work by accident (they are there for `d/dt(...)`).

The `print(x)` is also a stray debug statement: it dumps the offending
expression into the middle of the progress bar (`exp(lT)` above), which is what
makes the failure look like corrupted console output during a fit.

## Behaviour depends on the model's line count

`.rxDisguiseCmt()` already handles this correctly for the chunked path -- it hex-
encodes the whole `past(...)` left-hand side so the duration is never parsed as
an lhs. But chunking only kicks in above `chunkLines` (default 40), so the same
model optimizes or fails purely on its length:

```r
base <- c("lT=2.549445","a=1","b=0.5","k3=5","k1=0.535","k2=0.352",
          "G(0)=a","d/dt(G)=k3-(k1/k2)*(1-exp(-k2*t))*G",
          "past(G,exp(lT))=a*exp(b*t)","gdel=delay(G,exp(lT))")

rxOptExpr(paste(base, collapse = "\n"))                       # 10 lines -> error
rxOptExpr(paste(c(paste0("pad", 1:40, "=", 1:40, "*a+k1*k2"), # 50 lines -> ok
                  base), collapse = "\n"))
```

The long form round-trips the `past()` line back byte-for-byte and the optimized
model solves identically to the original, so the chunked path is the correct
behaviour and the whole-model path is the one that needs fixing.

## Suggested fix

Do not treat the `past()` duration as a left-hand side. Either emit it verbatim
(`deparse1(x[[3]])`) in `..rxOptLhs()`, or route it through `..rxOpt()` so it is
optimized like an ordinary right-hand-side expression. Either way the stray
`print(x)` at `R/rxOptExpr.R:178` should go.

## Session

rxode2 5.1.7, R 4.6, Linux.

### LINKED ISSUE #1193: past() duration is not resolved when building estimation models: generated DDE model fails .rxValidatePast() ('duration T does not match any delay(G, ...)')
When rxode2 builds an estimation model for a DDE, the `past()` line it emits
keeps the delay duration as the **verbatim source text** from the user's model,
while every `delay()` in the same model has had its duration inlined by
symengine. The two no longer agree, so `rxSolve()` rejects the model rxode2
itself just generated.

For nlmixr2's Example 6 DDE model (`T <- exp(tT + eta.T)`, `past(G, T) <- a*exp(b*t)`)
the generated FOCEi inner model contains:

```
d/dt(I)=G*exp(THETA[1])-exp(THETA[1])*delay(G, exp(ETA[1]+THETA[3]))
d/dt(D)=-D*exp(THETA[2])+exp(THETA[1])*delay(G, exp(ETA[1]+THETA[3]))
past(G,T)=1*exp(t*exp(THETA[4]))
```

`T` is neither a parameter nor an lhs of that model -- it is a dangling symbol.
Solving it fails:

```
Error : past(G, T): duration 'T' does not match any delay(G, ...) (found: exp(ETA[1] + THETA[3]))
```

## Reproducible example

```r
library(rxode2)
library(nlmixr2est)

mod6 <- function() {
  ini({
    tk4 <- log(0.15); tk5 <- log(0.10); tT <- log(11); tb <- log(0.40)
    eta.T ~ 0.0782
    prop.pk <- 0.382; add.tas <- 1.49; add.aks <- 0.420
  })
  model({
    k <- 0.184; V <- 0.039; k1 <- 0.535; k2 <- 0.352; k3 <- 5
    k4 <- exp(tk4); k5 <- exp(tk5); Emax <- 1; EC50 <- 44.3
    T  <- exp(tT + eta.T); I0 <- 2.42; a <- 1; b <- exp(tb)
    d/dt(Ac) <- -k * Ac
    c   <- Ac / V
    eff <- (Emax * c) / (EC50 + c)
    G(0) <- a
    d/dt(G) <- k3 - eff * G - (k1 / k2) * (1 - exp(-k2 * t)) * G
    past(G, T) <- a * exp(b * t)          # <- non-constant pre-history
    I(0) <- I0
    d/dt(I) <- k4 * G - k4 * delay(G, T)
    D(0) <- 0
    d/dt(D) <- k4 * delay(G, T) - k5 * D
    TAS <- I + D
    AKS <- D
    c ~ prop(prop.pk); TAS ~ add(add.tas); AKS ~ add(add.aks)
  })
}

## capture the model text nlmixr2est hands to rxode2()
txt <- NULL
.o <- nlmixr2est:::.nlmixr2estRxode2
assignInNamespace(".nlmixr2estRxode2", function(model, role, ...) {
  if (role == "rxInner" && is.null(txt)) txt <<- model
  .o(model, role, ...)
}, ns = "nlmixr2est")

fit <- nlmixr2(mod6, dat6, est = "focei",
               control = foceiControl(covMethod = "", maxOuterIterations = 0L))

grep("past\\(|delay\\(", strsplit(txt, "\n")[[1]], value = TRUE)
#> "d/dt(I)=G*exp(THETA[1])-exp(THETA[1])*delay(G, exp(ETA[1]+THETA[3]))"
#> "d/dt(D)=-D*exp(THETA[2])+exp(THETA[1])*delay(G, exp(ETA[1]+THETA[3]))"
#> "past(G,T)=1*exp(t*exp(THETA[4]))"                                  # <- 'T', not exp(ETA[1]+THETA[3])

## the generated model cannot be solved
p <- c(`THETA[1]`=log(.15), `THETA[2]`=log(.1), `THETA[3]`=log(11), `THETA[4]`=log(.4),
       `THETA[5]`=.382, `THETA[6]`=1.49, `THETA[7]`=.42, `ETA[1]`=0.2, CMT=6)
ev <- et(seq(0, 28, by = 0.5)); for (td in c(1, 8, 15)) ev <- et(ev, amt = 1, cmt = 1, time = td)
rxSolve(rxode2(txt), p, ev, method = "dop853", dense = TRUE)
#> Error : past(G, T): duration 'T' does not match any delay(G, ...) (found: exp(ETA[1] + THETA[3]))
```

## Cause

`R/symengine.R:1182` stores the duration as raw source text:

```r
assign(paste0("rx__pastTau_", .pastState, "__"),
       paste(deparse(x[[2]][[3]]), collapse = " "), envir = envir)
```

All three consumers then resolve the history **RHS** through the symengine env
but pass that raw `.tauTxt` straight through:

* `.rxPastBaseLinesFromEnv()` -- `R/dde.R:89` / `:96`
* `.rxDelaySensAugment()` -- `R/dde.R:386` / `:394`
* `.rxDelaySensAugment2()` -- `R/dde.R:851` / `:870`

e.g. `R/dde.R:91-96`:

```r
## resolve through the env so the injected line references root parameters
.rhsB <- tryCatch(eval(parse(text = .rhsTxt), envir = model), error = function(e) NULL)
.rhsOut <- if (!is.null(.rhsB) && inherits(.rhsB, "Basic")) rxFromSE(.rhsB) else .rhsTxt
.lines <- c(.lines, sprintf("past(%s,%s)=%s", .si, .tauTxt, .rhsOut))
##                                  ^^^^^^^ never resolved
```

The comment states the intent ("resolve through the env so the injected line
references root parameters") -- it just was never applied to the duration.

## Scope of the damage

**No runtime consequence.** The `past()` duration is captured purely as source
text for round-tripping and R-side matching (`src/parseCmtProperties.h:94-99`)
and is never used in the generated C: `_rxDelay()` calls
`_rxPast(_ind->id, i, td, ...)` whenever `td <= _ind->delayT0`, whatever the
duration says (`src/codegen2.h:396-397`). `T` never becomes a model parameter.
I confirmed this directly -- with `.rxValidatePast()` stubbed out, a mismatched
and a matched `past()` duration give bit-identical solutions
(`max|diff| = 0`), and the fit's OBJF is the same either way.

So fits are **not** silently wrong. What breaks is that the machine-generated
models cannot be round-tripped through `rxSolve()` -- they only run because
nlmixr2est solves them through its own path, which never reaches
`.rxValidatePast()` (`R/rxsolve.R:2600`).

Note that `.rxValidatePast()` already skips machine-generated `rx__sens_*`
histories as "trusted"; the base state's history is equally machine-generated
here but is not skipped, which is what surfaces the mismatch.

## Fix (confirmed)

Resolve the duration through the model env exactly as the RHS already is. I
tested this by patching `.rxDelaySensAugment()` to normalize
`rx__pastTau_<state>__` through the env before it runs:

```r
.b <- tryCatch(eval(parse(text = .tauTxt), envir = model), error = function(e) NULL)
if (!is.null(.b) && inherits(.b, "Basic")) .tauTxt <- rxFromSE(.b)
```

The generated line becomes

```
past(G,exp(ETA[1]+THETA[3]))=1*exp(t*exp(THETA[4]))
```

which matches the `delay()` terms, `rxSolve()` accepts the generated model, and
the fit gives the identical OBJF (552.7717) as before. The same change is needed
at all three sites.

## Session

rxode2 5.1.7, nlmixr2est 7.0.2, R 4.6, Linux.

mattfidler:
Closes #1192
Closes #1193

Two defects in how the delay duration of a `past(state, tau) <- expr` line is
carried through the model machinery.  Together they are why the nlmixr2 DDE
vignette has to pass `optExpression = FALSE` to every fit.

## #1192 -- `rxOptExpr()` could not optimize an expression duration

`..rxOptLhs()` recursed into the duration with the *left-hand-side* renderer,
which only knows names and the dosing-property heads (`/` and `(` work only by
accident -- they are there for `d/dt(...)`).  So `past(G, exp(lT))`,
`past(G, tau*2)` and `past(G, exp(THETA[1]+ETA[1]))` all hit
`stop("unsupported lhs in optimize expression")`, after a stray `print()` dumped
the duration into the middle of the progress bar.

Only the whole-model path was affected -- `.rxDisguiseCmt()` hex-encodes the
whole left-hand side for the chunked path -- so the same model failed at 10
lines and optimized at 50.  End to end:

```
optExpression = TRUE  -> Error : stopped optimizing duplicate expressions
optExpression = FALSE -> runs
```

The duration is now rendered by `.rxOptExpr()`, the right-hand-side optimizer.
It parses, and it picks up the *same* `rx_expr_` temporary the matching
`delay()` calls do -- which is what `.rxValidatePast()` needs to keep matching
them.  A bare name, a number and `(x)` still render byte-identically.

The chunked path cannot do that (the left-hand side is hidden from the optimizer
and restored byte-exactly), so `.rxRealignPastTau()` re-points a restored
`past()` duration at the one its `delay()` calls ended up with.  It is text only
-- the duration is never evaluated, it exists so a history can be matched to its
`delay()` (`src/parseCmtProperties.h`) -- and it never guesses: optimizing does
not reorder statements, so the pre-optimization text says which optimized
duration each `past()` line meant; a single duration is unambiguous without it;
anything else is left for the validator to report.

## #1193 -- the duration was never resolved in generated models

`rx__pastTau_<state>__` is stored as verbatim source text.  All four `past()`
emitters resolved the history *right-hand side* through the symengine env but
passed the duration straight through, while every `delay()` had its duration
inlined.  A duration written as an intermediate was therefore emitted into a
generated model that intermediate had been dead-code eliminated from:

```
d/dt(I)=G*exp(THETA[1])-exp(THETA[1])*delay(G, exp(ETA[1]+THETA[3]))
past(G,T)=1*exp(t*exp(THETA[4]))          # T is a dangling symbol
```

`rxSolve()` then rejected rxode2's own generated model with
`past(G, T): duration 'T' does not match any delay(G, ...)`.  Reachable from
rxode2 alone via `rxode2(mod, calcJac = TRUE)` / `calcSens=`, not only through
nlmixr2.

`.rxPastFromEnv()` / `.rxTauFromEnv()` now resolve the duration the way the
history already was, by evaluating a surrogate `delay(state, tau)` in the env so
the rendering is identical to the augmented `d/dt()` by construction (this also
covers a constant-folded duration, stored as a plain numeric rather than a
`Basic`).  Used at all four sites, including the exported
`.rxPastBaseLinesFromEnv()` -- its signature is unchanged, so released reverse
dependencies simply get a corrected string.

**No runtime behaviour changed.** The duration reaches only `sbt` (the
normalized-text buffer), never `sb`/`sbDt`, so it is absent from the generated C
entirely; `_rxDelay()` calls `_rxPast()` for any pre-history time regardless.
With `.rxValidatePast()` stubbed out, a matched and a mismatched duration give
bit-identical solutions.  Fits were never silently wrong -- the generated models
just could not be round-tripped through `rxSolve()`.

## Verification

- Duration matrix on the whole-model path: `exp(lT)`, `lT*2`, `2^lT`, `2*3`,
  `exp(THETA[1]+ETA[1])` all optimize (all errored before); `lT`, `12.8`, `(lT)`
  unchanged.
- Both paths keep the pairing (`past(G,rx_expr_0)` + `delay(G, rx_expr_0)`;
  chunked `rx_expr_c2_1`), and the optimized model solves identically.
- `calcJac=TRUE` / `calcSens=` emit `past(y,exp(lT))`, pass validation, and solve
  matching the plain model.
- nlmixr2 Example 6 (the vignette model): OBJF 552.7717 with `optExpression`
  both `TRUE` and `FALSE`, and `rxSolve()` now accepts the generated FOCEi inner
  model in **both** cases (it errored in both before), identical `D[20]`.
- A genuinely wrong duration is still rejected, with the original message, on
  both paths.

`test-opt-expr.R` 81 -> 112 assertions, `test-dde-past.R` 26 -> 49, both green
locally, as were `test-dde.R`, `test-dde-sens.R` and the adjoint suite (3230
assertions passing before the second commit's rewrite; the individual DDE and
optimizer files re-ran green after it).  Leaving the rest to CI.

The second commit is the result of an independent review pass, which found three
gaps in the first: a duration containing `=` split the line in the wrong place;
the delay scan parsed the reassembled model as R, which is rxode2 syntax and not
guaranteed to be valid R; and a state with several distinct delay durations was
skipped outright.  Both scans now match parentheses instead, the way
`.rxDisguiseDelayChunks()` already does.

## Follow-up (not in this PR)

`~/src/nlmixr2/vignettes/delays.Rmd` no longer needs `optExpression = FALSE`
once this ships.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,25 @@
 # rxode2 5.1.7
 
+## Bug fixes
+
+### Delay differential equations
+
+- `rxOptExpr()` no longer fails on a `past(state, tau)` whose delay duration is
+  an expression rather than a name or a number (`past(G, exp(lT))`,
+  `past(G, tau*2)`), which raised `unsupported lhs in optimize expression` and
+  printed the duration into the middle of the progress bar.  This made
+  `optExpression=TRUE` unusable for such a delay differential equation; it now
+  optimizes, and the duration follows the same common subexpression its
+  `delay()` terms do, so the history stays matched to them.
+
+- A generated delay differential equation model (`rxode2(..., calcJac=TRUE)`,
+  `calcSens=`, or an nlmixr2 estimation model) now resolves the `past()` delay
+  duration the same way it resolves the history itself.  A duration written as
+  an intermediate (`T <- exp(lT)`) was emitted verbatim while every `delay()`
+  had its duration inlined, so the generated model named a duration no `delay()`
+  used any more and `rxSolve()` rejected it with `duration 'T' does not match
+  any delay(...)`.
+
 # rxode2 5.1.6
 
 ## New features
--- a/NEWS.md
+++ b/NEWS.md
@@ -18,7 +18,10 @@
   an intermediate (`T <- exp(lT)`) was emitted verbatim while every `delay()`
   had its duration inlined, so the generated model named a duration no `delay()`
   used any more and `rxSolve()` rejected it with `duration 'T' does not match
-  any delay(...)`.
+  any delay(...)`.  This also covers a duration or a history written with
+  `THETA[n]`/`ETA[n]`, as every mu-referenced model is: they were left
+  unresolved, and an unresolved history additionally emitted no per-parameter
+  sensitivity pre-history at all.
 
 # rxode2 5.1.6
 
--- a/R/dde.R
+++ b/R/dde.R
@@ -93,9 +93,11 @@
 #' Resolve a stored past() delay duration through a symengine env
 #'
 #' The duration is rendered by evaluating a surrogate `delay(state, tau)` in the
-#' env and taking its second argument back off, so it comes out byte-identical to
-#' the duration `.rxDelaySensAugment()` reads off the augmented `d/dt()` (which is
-#' `deparse1()` of the same `rxFromSE()` round trip).  Evaluating the duration on
+#' env and taking its second argument back off, so it comes out as the same
+#' expression the duration `.rxDelaySensAugment()` reads off the augmented
+#' `d/dt()` is (both are `deparse1()` of the same `rxFromSE()` round trip; the
+#' spacing `deparse1()` puts around an operator is normalized away when a history
+#' is matched to its `delay()`).  Evaluating the duration on
 #' its own is the fallback; a constant-folded intermediate is stored as a plain R
 #' numeric rather than a `Basic`, which is exactly the case the surrogate gets
 #' right and a bare `eval()` does not.  An unresolvable duration (env binding
--- a/R/dde.R
+++ b/R/dde.R
@@ -67,6 +67,69 @@
   .df
 }
 
+#' Resolve a stored past() delay duration through a symengine env
+#'
+#' The duration is rendered by evaluating a surrogate `delay(state, tau)` in the
+#' env and taking its second argument back off, so it comes out byte-identical to
+#' the duration `.rxDelaySensAugment()` reads off the augmented `d/dt()` (which is
+#' `deparse1()` of the same `rxFromSE()` round trip).  Evaluating the duration on
+#' its own is the fallback; a constant-folded intermediate is stored as a plain R
+#' numeric rather than a `Basic`, which is exactly the case the surrogate gets
+#' right and a bare `eval()` does not.  An unresolvable duration (env binding
+#' missing, eval error) keeps its stored text.
+#'
+#' @param model symengine environment.
+#' @param state state the history belongs to.
+#' @param tauTxt stored duration text (`rx__pastTau_STATE__`).
+#' @return duration text to emit.
+#' @noRd
+.rxTauFromEnv <- function(model, state, tauTxt) {
+  if (is.null(tauTxt)) return(tauTxt)
+  .b <- tryCatch(eval(parse(text = paste0("delay(", state, ",", tauTxt, ")")),
+                      envir = model), error = function(e) NULL)
+  if (inherits(.b, "Basic")) {
+    .c <- tryCatch(parse(text = rxFromSE(.b))[[1L]], error = function(e) NULL)
+    if (is.call(.c) && length(.c) == 3L) return(deparse1(.c[[3L]]))
+  }
+  .t <- tryCatch(eval(parse(text = tauTxt), envir = model), error = function(e) NULL)
+  if (inherits(.t, "Basic")) return(rxFromSE(.t))
+  tauTxt
+}
+
+#' Read a state's stored past() history out of a symengine env, resolved
+#'
+#' Both the history RHS and the duration are stored as plain rxode2 text
+#' (`rx__pastRhs_STATE__` / `rx__pastTau_STATE__`, see `R/symengine.R`).  The RHS
+#' has always been resolved back through the env; the duration was not, so a
+#' duration written as an intermediate (`T`) was emitted verbatim even though that
+#' intermediate had been dead-code eliminated from the generated model, while the
+#' matching `delay()` had its duration inlined by symengine.  Resolving both here
+#' is what keeps the emitted `past()` line matched to its `delay()` terms, and
+#' keeps the three emitters from drifting apart.
+#'
+#' @param model symengine environment.
+#' @param state state the history belongs to.
+#' @return `NULL` when the state has no `past()`, otherwise a list with the
+#'   resolved duration `tau`, the resolved history text `rhs`, and the history
+#'   `Basic` `rhsB` (`NULL` when it did not resolve) for differentiation.
+#' @noRd
+.rxPastFromEnv <- function(model, state) {
+  .rhsTxt <- base::mget(paste0("rx__pastRhs_", state, "__"), envir = model,
+                        ifnotfound = list(NULL))[[1L]]
+  if (is.null(.rhsTxt)) return(NULL)
+  .tauTxt <- base::mget(paste0("rx__pastTau_", state, "__"), envir = model,
+                        ifnotfound = list(NULL))[[1L]]
+  ## resolve the duration first: rxFromSE() of a Basic holding lag0()/llik*()
+  ## poisons the next `[[`/get read from the env, and the history RHS is the one
+  ## that may carry such a call
+  .tau <- .rxTauFromEnv(model, state, .tauTxt)
+  .rhsB <- tryCatch(eval(parse(text = .rhsTxt), envir = model),
+                    error = function(e) NULL)
+  list(tau = .tau,
+       rhs = if (inherits(.rhsB, "Basic")) rxFromSE(.rhsB) else .rhsTxt,
+       rhsB = if (inherits(.rhsB, "Basic")) .rhsB else NULL)
+}
+
 #' Base past(state, tau) <- expr history lines from a symengine env
 #'
 #' Rebuilds the base `past(state,tau)=expr` line(s) from the stored
@@ -83,16 +146,10 @@
   .states <- tryCatch(rxode2::rxStateOde(model), error = function(e) character(0))
   .lines <- character(0)
   for (.si in .states) {
-    .rhsTxt <- base::mget(paste0("rx__pastRhs_", .si, "__"), envir = model,
-                          ifnotfound = list(NULL))[[1L]]
-    if (is.null(.rhsTxt)) next
-    .tauTxt <- base::mget(paste0("rx__pastTau_", .si, "__"), envir = model,
-                          ifnotfound = list(NULL))[[1L]]
     ## resolve through the env so the injected line references root parameters
-    .rhsB <- tryCatch(eval(parse(text = .rhsTxt), envir = model),
-                      error = function(e) NULL)
-    .rhsOut <- if (!is.null(.rhsB) && inherits(.rhsB, "Basic")) rxFromSE(.rhsB) else .rhsTxt
-    .lines <- c(.lines, sprintf("past(%s,%s)=%s", .si, .tauTxt, .rhsOut))
+    .p <- .rxPastFromEnv(model, .si)
+    if (is.null(.p)) next
+    .lines <- c(.lines, sprintf("past(%s,%s)=%s", .si, .p$tau, .p$rhs))
   }
   if (length(.lines)) .lines else NULL
 }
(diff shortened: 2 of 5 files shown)


mattfidler:
## Follow-up: CodeFactor, and seven rounds of independent review

CodeFactor is green (0 issues): the explicit `return()` on the new `past()` branch of
`..rxOptLhs()` is gone, and the new tests call `.rxValidatePast()` /
`.rxRealignPastTau()` / `..rxOptLhs()` unqualified, the way the rest of the suite does,
rather than through `rxode2:::`.

The branch was then reviewed by two outside models (Gemini via Antigravity, GPT via
Copilot), alternating until both came back clean. Every finding below was reproduced
before it was fixed, and each fix has a test. Three of the rounds were about code the
review rounds themselves had introduced.

### Wrong answers

- **A dropped duration was guessed at.** Optimizing can *remove* a `delay()` --
  `0*delay(G,10)` folds to `0` -- and the single-duration fallback then re-pointed
  `past(G,10)` at the surviving `delay(G,20)`. A model `.rxValidatePast()` should reject
  was accepted and **solved with one delay term carrying another's history**. A state's
  durations are now matched only when both texts agree on how many it has; the fallback
  is gone and nothing is guessed at.

- **`THETA[]`/`ETA[]` were never resolved.** The stored `past()` pieces are rxode2 source
  text, where `THETA[1]` is a subscript; in the symengine env it is the symbol `THETA_1_`,
  so evaluating the text directly failed on every mu-referenced model. `.rxTauFromEnv()`
  then fell back to the verbatim duration while the matching `delay()` had its own
  inlined -- the #1193 mismatch, still reachable from
  `rxode2(mod, calcJac = TRUE)` with a `THETA[]` duration. The same eval resolved the
  history, and an unresolved history is not a `Basic`, so `.rxDelaySensAugment()` also
  emitted **no per-parameter sensitivity pre-history at all**. Both now go through
  `.rxSeEvalTxt()`, which translates with `.rxToSE()` first.

### Models wrongly accepted

- `.rxRealignPastTau()` re-pointed a duration that had matched no `delay()` *before*
  optimizing, so the chunked path silently accepted `past(G,typo)` by rewriting it to the
  optimizer's temporary -- while the whole-model path rejects it. It now rewrites only a
  duration that did match one. (Two later rounds closed the same hole again via the
  optional `orig` argument and an unreadable pre-optimization text; `orig` is now
  required, and an unreadable text concludes nothing.)

- Two histories on one state could both be rewritten to the same duration, leaving one
  silently shadowing the other.

### Models wrongly rejected

- `.rxSplitStateTau()` required a state name to start with a letter, so a leading-dot
  state (`d/dt(.y.1)` parses) had its `past()` line skipped.
- A `delay()` written in a `#` comment gave the state a duration no call carries, and the
  realign then refused the rewrite the `past()` line needed.
- A `delay()` written in a *string* (`s = "delay(G,"` -- a model that parses) marked the
  text unreadable and stopped the realign for all of it.
- ... and then masking strings corrupted a duration that legitimately holds one
  (`delay(G, 1+(OCC=="first"))`). A call is now **located** on the code-only form of a
  line and **read** off the line itself, the top-level comma likewise, so a comma or
  parenthesis inside a string cannot split it. `.rxCodeOnly()` blanks rather than removes
  precisely so the two forms agree column for column.

### Considered, deliberately unchanged

- An unreadable `delay()` (one split over lines) stops the realign for *every* state, not
  just its own. That is the safe direction, and such a model errors exactly as it did
  before this pass existed.
- `deparse1()` spaces an operator duration out (`past(G,a * b)` against `delay(G,a*b)`).
  Matching normalizes that away; a test now solves such a model over its whole history
  window and gets the reference answer.
- With `T = exp(lT)` and `U = exp(lT)`, `past(y,U)` against `delay(y,T)` is rejected as
  written but accepted once differentiated, because the generated model resolves both to
  `exp(lT)`. Resolving is the point of #1193, the two durations *are* the same
  expression, and the model solves identically to `past(y,T)`. Making the source-model
  check resolve durations too would change a check on a path every reverse dependency
  executes, so I have left it alone rather than decide that here.

### Tests

`test-opt-expr.R` and `test-dde-past.R` are at 311 assertions between them (from 107 on
`main`), all green, and the full local suite is **0 failures / 0 errors / 73,981 passing**.

Generated with [Claude Code](https://claude.com/claude-code)

PR #1198, opened 2026-08-07
fix(ui): escape quotes/backslashes when promoting ini() comments to label()

### LINKED ISSUE #1195: `ini()` comment-to-`label()` promotion breaks on a `"` (or `\`) in the comment: `sub()` strips `deparse1()`'s escaping
## Summary

When a model function is parsed **with source refs retained**, `rxode2:::.rxReplaceCommentWithLabel()` promotes a trailing `#` comment on an `ini({})` line into a `label("...")` call. If the comment text contains a double quote, the generated `label()` string terminates early and the subsequent re-parse fails with a syntax error. A lone backslash in the comment fails the same way (`'\L' is an unrecognized escape`).

The label text *is* escaped correctly by `deparse1()` — the escaping is then undone one line later, because the escaped string is interpolated into the **replacement** argument of `sub()`, and `sub()` treats backslashes in a replacement as escapes and strips one level.

Any embedded double quote triggers it, balanced or not — the generated label is itself delimited by double quotes.

## Reprex

```r
## rxode2 reprex: a `"` in a trailing ini() comment breaks label() promotion
library(rxode2)

mod <- tempfile(fileext = ".R")
writeLines(c(
  'f <- function() {',
  '  ini({',
  '    lka <- 0.5   # fixed to a "small value"',
  '    addSd <- 1',
  '  })',
  '  model({',
  '    ka <- exp(lka)',
  '    d/dt(central) <- -ka * central',
  '    Cc <- central',
  '    Cc ~ add(addSd)',
  '  })',
  '}'), mod)

## (1) WITH source refs -> comment is promoted to label() -> parse error
withSrc <- new.env()
source(mod, local = withSrc, keep.source = TRUE)
print(try(rxode2::rxode(withSrc$f), silent = TRUE))

## (2) WITHOUT source refs -> no promotion -> same model resolves fine
noSrc <- new.env()
sys.source(mod, envir = noSrc, keep.source = FALSE)
cat("no-srcref path:", class(rxode2::rxode(noSrc$f))[1], "\n")
```

Output (rxode2 5.1.3, R 4.6.1):

```
ℹ parameter labels from comments will be replaced by 'label()'
<simpleError in parse(text = paste(.ret, collapse = "\n"), keep.source = FALSE): <text>:3:39: unexpected symbol
2:   ini({
3:     lka <- 0.5   ; label("fixed to a "small
                                         ^>

ℹ parameter labels from comments are typically ignored in non-interactive mode
ℹ Need to run with the source intact to parse comments
no-srcref path: rxUi
```

Note the generated line in the error message: `label("fixed to a "small value"")`.

`Rscript` sets `keep.source = FALSE` by default, so the `keep.source = TRUE` argument to `source()` above is what makes the failing path reachable in a non-interactive run.

## Traceback

```
 7: nlmixr2est::nlmixr(e$f)
 8: nlmixr2.function(e$f)
 9: rxode2::rxode2(object)
10: rxUiDecompress(.rxFunction2ui(model))
11: .rxFunction2ui(model)
12: .rxFunctionRearrange(eval(parse(text = paste(.rxFunction2string(fun),
15: eval(parse(text = paste(.rxFunction2string(fun), collapse = "\n")))
16: parse(text = paste(.rxFunction2string(fun), collapse = "\n"))
18: .rxFunction2string(fun)
19: .rxReplaceCommentWithLabel(as.character(.srcRef, useSource = TRUE))
20: deparse(eval(parse(text = paste(.ret, collapse = "\n"), keep.source = FALSE)))
22: parse(text = paste(.ret, collapse = "\n"), keep.source = FALSE)   ## <- error
```

`rxode2::rxode()` alone produces the same failure; `nlmixr2est::nlmixr()` is just how it surfaced for us.

## Root cause

In `rxode2:::.rxReplaceCommentWithLabel()`:

```r
.label <- deparse1(sub(.regLabel, "\\2", line))
return(sub(.regLabel, paste0("\\1; label(", .label, ")"), line))
```

`deparse1()` produces a properly escaped literal, but `sub()` then parses backslashes in the replacement string and removes one level:

```r
line <- '    lka <- 0.5   # fixed to a "small value"'
.regLabel <- "^( *[^\n\"]+) *#+ *(.*) *$"

deparse1(sub(.regLabel, "\\2", line))
#> [1] "\"fixed to a \\\"small value\\\"\""      # correctly escaped

sub(.regLabel, paste0("\\1; label(", deparse1(sub(.regLabel, "\\2", line)), ")"), line)
#> [1] "    lka <- 0.5   ; label(\"fixed to a \"small value\"\")"   # escaping gone
```

This also explains the backslash case (`mg\L` becomes an invalid `\L` escape) and why an *already*-escaped `\"` in the comment happens to survive — `sub()` strips it back down to a valid `\"`.

## Boundary matrix

Reproduced end-to-end against rxode2 5.1.3 / R 4.6.1. `srcref` = `source(..., keep.source = TRUE)`; `no srcref` = `sys.source(..., keep.source = FALSE)`.

| # | Case | srcref | no srcref |
|---|------|--------|-----------|
| 01 | `<-` line, no `label()`, plain comment (control) | OK | OK |
| 02 | `<-` line, no `label()`, comment contains `"` | **FAIL** | OK |
| 03 | `<-` line, already has `label()`, comment contains `"` | OK | OK |
| 04 | `~` eta line, no `label()`, plain comment (control) | OK | OK |
| 05 | `~` eta line, no `label()`, comment contains `"` | **FAIL** | OK |
| 06 | `~` eta line, already has `label()`, comment contains `"` | OK | OK |
| 07 | `~ fixed()` eta, no `label()`, comment contains `"` | **FAIL** | OK |
| 08 | comment contains `'` (single quote) instead | OK | OK |
| 09 | comment contains an escaped `\"` | OK | OK |
| 10 | comment contains a lone backslash (`mg\L`) | **FAIL** | OK |
| 11 | standalone `#` comment line inside `ini({})` containing `"` | OK | OK |
| 12 | comment containing `"` inside `model({})` rather than `ini({})` | OK | OK |
| 13 | comment contains a single *unbalanced* `"` (`6" tall`) | **FAIL** | OK |

So the trigger is precisely: **an `ini({})` line that has a trailing comment, has no pre-existing `label()`, and whose comment text contains `"` or `\`.** Both `<-` and `~` lines are affected.

Cases 03/06 are safe because `.regLabel`'s first group is `[^\n\"]+`, which cannot span a `"` — a line whose code part already contains a quoted string never matches, so no promotion is attempted. (Side effect worth noting: the trailing comment on such a line is then silently discarded rather than becoming a label. That is arguably correct here, since an explicit `label()` should win, but it is worth being deliberate about.)

## Why this is worth fixing rather than documenting

The failure is **srcref-conditional**, which makes it confusing in practice:

- resolve the model without source refs (`sys.source()`; `R CMD build`/`INSTALL` with the default `keep.source = FALSE`) — works, because the promotion never runs;
- resolve the same model with source refs (`source(..., keep.source = TRUE)`, or an interactive session, or a test suite run with `keep.source = TRUE`) — syntax error.

So a package build can be green while its test suite is red on the identical model file. And because the error is raised by a bare `parse()` deep inside `.rxReplaceCommentWithLabel()`, the message reports a line number **in the regenerated text**, with no indication of which model function or which source line is responsible. In a package with many model files this is quite hard to localise.

We hit this in [`nlmixr2lib`](https://github.com/nlmixr2/nlmixr2lib) on a line that quoted the source paper verbatim:

```r
etalki50 ~ fixed(0.0225);  # Table 4: IIV KI50 variance = 0.0225 (fix) -- Online Resource 1 Sect. II: "fixed to a small value (0.0225) to enhance the numerical estimation capability"
```

For calibration on how rare-but-silent this is: scanning all 1705 models in that library for the shape turned up exactly one instance.

## Suggested fix

Build the output by concatenation instead of routing the already-escaped label through `sub()`'s replacement parser:

```r
.env$convertLabel <- TRUE
.label <- deparse1(sub(.regLabel, "\\2", line))
return(paste0(sub(.regLabel, "\\1", line), "; label(", .label, ")"))
```

I checked this two ways. At the string level it is byte-identical to the current output on every case that currently works (including the single-quote case), and it fixes 02, 05, 07, 10 and 13:

```
LINE     :     lka <- 0.5      # fixed to a "small value"
  current :     lka <- 0.5      ; label("fixed to a "small value"")     PARSE ERROR
  proposed:     lka <- 0.5      ; label("fixed to a \"small value\"")   parses OK
            recovered label -> [1] "fixed to a \"small value\""
```

End-to-end, patching the function into the namespace (`assignInNamespace`) and re-running all 13 cases through `rxode2::rxode()` on the srcref path gives 13/13 OK, with the comment text round-tripping intact into `iniDf$label`:

```
01-plain-no-label            OK  labels: ‘absorption rate’
02-dquote-no-label           OK  labels: ‘fixed to a "small value"’
03-dquote-with-label         OK  labels: ‘absorption’
04-eta-plain-no-label        OK  labels: ‘IIV on ka’
05-eta-dquote-no-label       OK  labels: ‘IIV "fixed" per the paper’
06-eta-dquote-with-label     OK  labels: ‘IIV ka’
07-eta-fixed-dquote          OK  labels: ‘IIV variance "fixed to a small value (0.0225)"’
08-single-quote              OK  labels: ‘fixed to a 'small value'’
09-escaped-dquote            OK  labels: ‘fixed to a \"small value\"’
10-backslash                 OK  labels: ‘units are mg\L’
11-standalone-comment        OK  labels:
12-dquote-in-model           OK  labels:
13-one-dquote                OK  labels: ‘6" tall’
```

An equivalent minimal alternative, if you would rather keep the single `sub()` call, is to double the backslashes in the replacement first — `.label <- gsub("\\", "\\\\", .label, fixed = TRUE)` — but the `paste0()` form avoids the class of bug entirely rather than compensating for it.

Two optional extras, both independent of the above:

1. Wrap the final `parse()` in `.rxReplaceCommentWithLabel()` so a failure reports the offending source line rather than a line number in the regenerated text. That is what cost us the most time here.
2. Consider whether the comment on a line that already has an explicit `label()` (cases 03/06) should be silently dropped or warned about.

## Session info

```
rxode2      5.1.3
rxode2ll    2.0.14
nlmixr2est  6.0.2
lotri       1.0.5
R version 4.6.1 (2026-06-24)
Platform: x86_64-pc-linux-gnu
Running under: Ubuntu 24.04.4 LTS
```

billdenney:
A trailing `#` comment on an `ini({})` line is promoted to a `label()` call by `.rxReplaceCommentWithLabel()` when the model is parsed with its source refs intact.  The label text was escaped correctly by `deparse1()`, but was then interpolated into the *replacement* argument of `sub()`, which parses backslashes in a replacement and strips one level.  That undid the escaping, so a comment containing a double quote generated `label("fixed to a "small value"")`, which failed the re-parse with a bare syntax error pointing into regenerated text rather than at the offending source line.  A lone backslash in the comment failed the same way.

Build the promoted line with `paste0()` instead, leaving `deparse1()`'s escaping intact.  Output is byte-identical for every comment that already worked, including ones containing single quotes.

Because the promotion only runs when source refs are kept, the same model resolved fine without them -- so a package build could be green while a test suite run with `keep.source = TRUE` was red on the identical file.

Adds an enumerating test over plain, double-quoted, unbalanced-quote, single-quoted, backslash, escaped-quote and mixed comments asserting the comment text round-trips into the label verbatim, plus exact-output checks for a population parameter and an eta line, and a check that a line already carrying `label()` is left alone.

Fixes #1195


mattfidler:
## Merged with `main`, reviewed by Copilot and Antigravity

`origin/main` merged in (`60c5177`). The only conflict was in `NEWS.md`, where this branch's bug-fix bullet and main's new "Delay differential equations" section landed at the same seam. Resolved as a union; main's third `## Bug fixes` heading was folded into the existing one so the DDE entries are a `###` subsection, per the NEWS conventions in `CLAUDE.md`. PR is `MERGEABLE` again.

### Review

Two independent reviewers ran over the merged diff: **GitHub Copilot** (`gpt-5.4`) and **Antigravity** (`gemini-3.1-pro-high`). Both agreed the `paste0()` rewrite is correct and behavior-preserving. I verified every claim against the code rather than relaying it.

I confirmed old/new output is byte-identical on trailing whitespace, `##`, a `#` inside the comment, an empty comment, and a 2000-character comment; and that `deparse1()` never splits a long string across elements.

**Confirmed -- missing coverage for the *silent* failure mode** (raised by both reviewers). The existing cases all failed *loudly* under the old code. Three did not:

| comment | old `label()` text | new |
|---|---|---|
| `# line1\nline2` | real newline | `line1\nline2` |
| `# a\tb` | real tab | `a\tb` |
| `# see \1 here` | `\001` | `see \1 here` |

Each re-parsed cleanly and carried the wrong text, so no test would have caught a regression. Added as `escapeN` / `escapeT` / `backReference`, plus `trailingSlash` (`# comment \`) at Antigravity's suggestion. `ae74038`, `bbe4793`.

**Rejected -- Antigravity's finding #1** ("`[^\n\"]` is a TRE bracket meaning *not backslash, n, or quote*, so `mean <- 0.45 # x` never promotes"). False. R's parser converts `\n` and `\"` in the string literal before the regex engine sees them; the bracket is `[^<LF>"]` and contains no backslash at all (char codes `91 94 10 34 93`). Verified: `mean <- 0.45` and `n_tka <- 0.45` both promote correctly.

**Out of scope, filed separately** -- both Antigravity (#2) and my own edge-case sweep found a real, silent, *pre-existing* bug in the same function: a comment containing `#` loses its label entirely, because group 1 is greedy and swallows the first `#`, leaving a stray `#` that comments out the generated `label()` call.

```r
tka <- 0.45 ## Log Ka          ->  tka <- 0.45 #; label("Log Ka")     # label dropped
tka <- 0.45 # rate # per hour  ->  tka <- 0.45 # rate ; label("per hour")  # label dropped
```

Identical before and after this PR, so not a regression. Tracked separately rather than widening this PR.

### CI

`test-coverage` is red, but **not because of this PR** -- see the separate note below.

mattfidler:
Follow-up on the red `test-coverage` check: it is pre-existing and not caused by this PR. It fails identically on `main` and on ~8 unrelated branches, and the tests themselves pass (`FAIL 0 | PASS 76505`).

The run dies afterwards in covr's merge -- covr writes a `covr_trace_*` from every R process that loads rxode2, and one truncated trace aborts `merge_coverage()`, which skips the codecov upload entirely. Diagnosed with the failure artifacts in #1207; fix in #1208.

Nothing to do on this branch.

PR #1202, opened 2026-08-07
fix: keep multi-row `params` rectangular when omega/sigma are NA

billdenney:
rxode2#1201

## The bug

`rxSolve()` given a multi-row `params` data.frame (one parameter set per `id`)
together with `omega = NA` returned **different results for byte-identical
inputs** -- some silently wrong, some non-finite -- once there were eight or more
subjects. `omega = NA` is documented to fix the between-subject etas at zero;
instead they were filled from adjacent heap memory.

Nothing surfaced it: no warning, no error, most values in each solve correct, and
the returned `$params` still reported the etas as zero.

## Root cause

In the `omega = NA` branch of `.rxSolveFromUi()`:

```r
params <- c(params, setNames(rep(0, dim(.omega)[1]), dimnames(.omega)[[2]]))
```

`c()` on a data.frame drops the `data.frame` class and returns a plain **ragged**
list -- the per-id columns keep length `nid`, the appended zeros have length one:

```r
pars <- data.frame(id = 1:4, lbase = c(.1,.2,.3,.4), lic50 = c(-1,-2,-3,-4))
lengths(c(pars, setNames(rep(0, 2), c("etalbase", "etalic50"))))
#>       id    lbase    lic50 etalbase etalic50
#>        4        4        4        1        1
```

The solver then reads `nid` values out of each length-1 vector, i.e. out of
bounds. Isolated by handing `rxSolve()` the list that line built versus the same
list with a full-length eta -- the only difference being the eta element's
*length*, not its class:

| `params` handed to `rxSolve()` | eta element | 50 solves, 16 ids |
|---|---|---|
| ragged list, as built by that line | length 1 | 1/50 wrong, 2 distinct |
| same list, full-length eta | length 16 | 0/50 wrong |
| data.frame with an eta column | length 16 | 0/50 wrong |

Introduced in cd6914784 (2022-04-01), first released in v2.0.6 -- long-standing,
not a regression.

## The fix

Adds an internal `.rxParamsZero(params, mat)` that assigns a column when `params`
is a data.frame, adds a column when it is a matrix, and falls back to `c()` for a
named numeric vector. The data.frame shape is the one `rxSolve()` already uses
for the same "treat these omega/sigma items as zero" operation further down the
same file, so it is the existing idiom rather than a new convention.

## Also fixed at the same code site

- **`sigma = NA` had the identical unguarded `c()`.** With a multi-row `params`
  data.frame it errored (`The following parameter(s) are required for solving:
  rxerr.base, addSd, tbase`); it now solves correctly.
- **A multi-row `params` matrix**, which `c()` destroys just as surely -- it
  drops the `dim` -- so `omega = NA`/`sigma = NA` failed outright with `The
  following parameter(s) are required for solving`. The zeros are added as
  columns.
- **`omega = NA` on a model with no between-subject variability.** `object$omega`
  is `NULL`, so `dim(NULL)[1]` is `NA` and `rep(0, NA)` raised
  `invalid 'times' argument` from inside `rxSolve()`.
- **`sigma = NA` on a model with no residual error.** `object$simulationSigma` is
  a 0x0 matrix there, so the old `c()` appended nothing but still dropped the
  data.frame class, failing with `The following parameter(s) are required for
  solving: tbase`.
- **A zero-row `params` data.frame.** Assigning a scalar column to one raises
  `replacement has 1 row, data has 0`; the zeros are recycled to the row count.

## And one more site with the same shape problem

An omega/sigma entry whose variance is zero (say `eta.base ~ fix(0)`) is dropped
from the matrix that is simulated from and recorded in `.ctl$.zeros`, to be
supplied to the model as a literal zero instead. `rxSolve.default()` did that for
a data.frame `params` and for a named numeric one, but a **matrix** `params`
matched neither branch, so the item never reached the solver and the solve failed
with `The following parameter(s) are required for solving: eta.base`. That branch
now goes through the same helper.

Deliberately conservative there: only the names the matrix is **missing** are
added. A matrix that already carried a column for a zeroed item had that column
used as given, and that keeps working. (A data.frame has always had such a column
overwritten with zero; this change does not touch that.)

## A matrix `params` was invisible to the rest of the solve setup too

Chasing the shape confusion further turned up two more, both silent and both
independent of the `c()` bug above.

**A supplied random effect in a matrix `params` was ignored and simulated over.**
`names()` of a matrix is `NULL` -- a matrix's names are its *column* names -- and
`rxSolve()` decides "did the caller already supply this random effect?" with
`v %in% names(params)`. For a matrix the answer was always no, so the effect
stayed in the `omega` and a random draw was used in place of the supplied column.
Same model, same `eta.base = 100` supplied for every subject, only the container
differs:

```
EXPECTED (tbase+100): 101 102 103 104 105 106
data.frame:           101 102 103 104 105 106
matrix    :          0.92 1.84 2.67 4.91 4.89 5.86
```

No warning, and the numbers look perfectly reasonable unless you know what they
should be. `.rxParamsNms()` asks each shape for its names the right way.

**Supplying one random effect stopped the others from being simulated.** A
supplied effect is dropped from the `omega` before solving, but that subset had
no `drop = FALSE`: taking a two-eta `omega` down to one gives a *scalar*, whose
`dim` is `NULL`, and `all(NULL == c(0L, 0L))` is `TRUE` -- so the whole `omega`
was discarded and the remaining eta was neither simulated nor supplied
(`The following parameter(s) are required for solving: eta.b`). The matching
`sigma` code already had the guard. This one is pre-existing for a data.frame
`params`; a matrix reaches that filter only now that its column names are
visible to it.

## `zeroVarParamHandle=`, and what happens to a value you supplied for a zeroed item

An omega/sigma item with zero variance is handed to the model as a literal zero.
If `params` already supplies a value for it, the shapes disagreed: a data.frame
had its value replaced by that zero, a matrix kept its own (only because the
matrix case was never written). Both replace it now, and the replacement no
longer happens silently:

- `zeroVarParamHandle = "warn"` (default) -- replace it with zero and say so.
- `"ignore"` -- replace it with zero silently.
- `"keep"` -- use the supplied value, which is the fit a matrix `params` gave
  before.

Also settable as `options(rxode2.zeroVarParamHandle=)`. The new control entry is
appended **last** in the `rxControl()` list, so every existing positional
`Rxc_*` index in `inst/include/rxode2parse_control.h` is unchanged, and a control
saved by an older rxode2 (which has no such element) reads as the default.

Worth knowing: through a `rxUi` model this corner case does not arise at all --
the `omega` filter in `.rxSolveFromUi()` drops a zero-variance item from the
`omega` when `params` supplies it, so it never reaches the zeroing code. It is
the classic `rxSolve(<rxode2 model>, params, omega = lotri(...))` path that gets
here.

## Tests

`test-ui-simulation.R`:

- `.rxParamsZero` unit test -- fully deterministic, pins the root cause directly
  (asserts the result stays rectangular, and covers the matrix, zero-row, vector
  and `NULL` paths).
- Integration test -- 200 subjects, 20 repeated solves, asserting the solved
  values equal the known-exact answer, for a data.frame and a matrix `params`.
  Deliberately asserts on the **solved values, not `$params`**: on the broken
  build `$params` reported the etas as zero in 10/10 runs while the solved values
  were wrong in 9/10. `sigma = NA` is asserted through `sim == base`, since the
  etas are still simulated there and `base` alone would not see the residual.
- The eta-free `omega = NA` and the residual-free `sigma = NA` no-ops.

`test-sim-zeros.R`: a zeroed omega item reaching a matrix `params`, the
data.frame control, and all three `zeroVarParamHandle` values across both shapes
(plus the nothing-supplied and `NULL`-handle cases).

`test-ui-simulation.R` also covers supplying one eta of two and asserting the
other still varies, for a data.frame and a matrix.

Confirmed to **fail on the unpatched build** (the multi-row cases hit testthat's
10-failure cap, with `NaN`, `250.0` and `442.0` in place of the expected values)
and pass with the fix.

## Verification

- `test-ui-simulation.R` + `test-sim-zeros.R`: 139 pass, 0 failures.
- Sweep of solve-related files (`test-par-solve.R`, `test-solver-basic.R`,
  `test-omega-chol.R`, `test-sim-zeros.R`, `test-pipeline.R`, `test-ui-solve.R`,
  `test-ui-simulation.R`): 5589 pass, 0 failures.
- The original 200-subject model that surfaced this becomes deterministic and
  exactly correct.

No documentation regeneration needed -- `.rxParamsZero()` is internal and
`@noRd`, so `NAMESPACE` and `man/` are untouched.

`origin/main` is merged in; the only conflict was `NEWS.md`, where the three
separate `## Bug fixes` headings that 5.1.7 had accumulated are consolidated into
one section with `###` subsections, per the repo's NEWS convention.


mattfidler:
Follow-up round: `origin/main` merged in, and three more fixes pushed
(`892f5ee5a`, `1291c466c`, `89bf7530b`). The PR description has been rewritten to
cover all of it; this is what changed since the first review pass.

### A matrix `params` was invisible to the rest of the solve setup

`names()` of a matrix is `NULL` -- a matrix's names are its *column* names -- and
`rxSolve()` decides "did the caller already supply this random effect?" with
`v %in% names(params)`. For a matrix the answer was always no, so the effect
stayed in the `omega` and a random draw was used in place of the supplied
column. Same model, `eta.base = 100` supplied for every subject, only the
container differs:

```
EXPECTED (tbase+100): 101 102 103 104 105 106
data.frame:           101 102 103 104 105 106
matrix (before):     0.92 1.84 2.67 4.91 4.89 5.86
matrix (after):       101 102 103 104 105 106
```

No warning, and the values look reasonable unless you know what they should be.
`.rxParamsNms()` asks each shape for its names the right way.

### That uncovered a third bug: supplying one eta stopped the others simulating

A supplied effect is dropped from the `omega` before solving, but that subset had
no `drop = FALSE`. Taking a two-eta `omega` down to one gives a **scalar**, whose
`dim` is `NULL`, and `all(NULL == c(0L, 0L))` is `TRUE` -- so the whole `omega`
was discarded and the remaining eta was neither simulated nor supplied
(`The following parameter(s) are required for solving: eta.b`). The matching
`sigma` code already had the guard. Pre-existing for a data.frame `params`; a
matrix reaches that filter only now that its column names are visible to it.

### `zeroVarParamHandle=` for the zero-variance corner case

An omega/sigma item with zero variance is handed to the model as a literal zero.
If `params` already supplies a value for it the shapes disagreed: a data.frame
had its value replaced, a matrix kept its own (only because the matrix case was
never written). Both replace it now, and no longer silently:

| | supplied value | warning |
|---|---|---|
| `"warn"` (default) | replaced by zero | yes, naming the items and both alternatives |
| `"ignore"` | replaced by zero | no |
| `"keep"` | used as given -- the fit a matrix `params` gave before | no |

Also settable as `options(rxode2.zeroVarParamHandle=)`. The new control entry is
appended **last** in the `rxControl()` list, so every existing positional
`Rxc_*` index in `inst/include/rxode2parse_control.h` is unchanged, and a control
saved by an older rxode2 (no such element) reads as the default.

Worth knowing: through a `rxUi` model this corner case does not arise at all --
the `omega` filter in `.rxSolveFromUi()` drops a zero-variance item when `params`
supplies it, so it never reaches the zeroing code. It is the classic
`rxSolve(<rxode2 model>, params, omega = lotri(...))` path that gets here.

### Review and verification

Reviewed again by GitHub Copilot (GPT) and Antigravity (Gemini) as independent
second opinions.

- Copilot: no findings on this round. It independently confirmed the positional
  `Rxc_*` claim and that no C code assumes the control-list length.
- Antigravity: flagged that the named-numeric-vector path appends a duplicate
  name, so the *supplied* value would win and contradict the warning. Checked it
  directly -- the appended zero wins (`base` = 1, not 6), so the warning is
  accurate and `"keep"` correctly gives 6. Rejected. Its other point, that the
  data.frame half of one test passes unpatched, is correct and deliberate: it is
  the control beside the matrix case.

`test-sim-zeros.R` + `test-ui-simulation.R`: 168 pass, 0 failures against the
final code. The eight solve-related files (`par-solve`, `solver-basic`,
`omega-chol`, `sim-zeros`, `pipeline`, `ui-solve`, `ui-simulation`, `rxsolve`):
5614 pass, 0 failures. Every new test was confirmed to fail on the unpatched
build. Two wider sweeps are still running on a heavily loaded machine; nothing in
them has failed.

One judgment call is still open, called out in the description: the classic-path
zeroing now replaces a supplied value for **both** shapes where a matrix used to
keep it. `zeroVarParamHandle="keep"` restores the old matrix fit, but the default
is a behavior change for that (narrow) case.

mattfidler:
CodeFactor is clean on this PR now: **52 issues fixed, 0 found** (`790db1fdb`).

The 12 it flagged were all in the two test files this PR touches:

| Count | Rule | Fix |
|---|---|---|
| 9 | `undesirable_operator_linter` -- `rxode2:::.rxParamsZero` / `rxode2:::.rxZeroVarParams` | bound once per test with `getFromNamespace()` |
| 2 | `unnecessary_concatenation_linter` -- `lotri(eta1~c(0.0))` | `lotri(eta1~0.0)` |
| 1 | `pipe_consistency_linter` -- `%>%` | `\|>`, which the rest of the file already uses |

The two `c(0.0)` ones are pre-existing lines (from `c752bbe60`) that the
re-analysis surfaced because the file changed. `identical(lotri(eta1~c(0.0)),
lotri(eta1~0.0))` is `TRUE`, checked before touching them, and `|>` is fine
against the declared `R (>= 4.1.0)`.

Rather than take the web report on faith, all 12 were reproduced locally by
running the three matching lintr linters against the pre-fix files -- they fired
at exactly the line numbers CodeFactor named (6 per file) -- and all three
linters are clean on both files afterwards.

### Verification

- `test-sim-zeros.R` + `test-ui-simulation.R`: 168 pass, 0 failures.
- The nine solve-related files (`par-solve`, `solver-basic`, `omega-chol`,
  `sim-zeros`, `pipeline`, `ui-solve`, `ui-simulation`, `rxsolve`, `control`):
  **5618 pass, 0 failures** against the final code.
- A wider `^ui|solve|sim|params|et$|event` sweep is still running; an earlier run
  of it reported one failure at `test-ui-simulation.R:926`, which was an artifact
  of editing the tree mid-run (that run had `load_all()`ed the package before the
  `drop = FALSE` fix and then picked the newly added test up off disk). The same
  test passes on the committed code in both runs above.

PR #1203, opened 2026-08-07
fix: raise an R condition, not a signal, when an Imports: package is missing

### LINKED ISSUE #1204: Missing `Imports:` package crashes R: SIGSEGV without rxode2ll, SIGABRT without cli
## Summary

Two independent ways that a missing `DESCRIPTION` `Imports:` package kills the R
process outright instead of raising an R condition:

| missing package | result |
|---|---|
| `rxode2ll` | **SIGSEGV** (exit 139) on the first `rxSolve()` of *any* compiled model |
| `cli` | **SIGABRT** (exit 134) inside `library(rxode2)` itself |

Neither is catchable -- `try()`, `tryCatch()` and testthat all die with the
process.

Both are on current `main` (v5.1.7, 7d74e29e4).

## Reproducible example

Run R with a library that has every rxode2 dependency except one. The crash needs
no special model -- a plain ODE with no `ini()`, no residual endpoint and no
`llik*()` call anywhere is enough:

```r
# with `rxode2ll` absent from .libPaths()
library(rxode2)
m <- rxode2("d/dt(x) <- -0.5*x;")
ev <- et(et(0:3), amt = 1, cmt = "x")
rxSolve(m, ev, cores = 1L)
```

```
Error in (function (x, local = TRUE, now = TRUE, ...)  :
  function 'rxLlikCauchyDscale' not provided by package 'rxode2ll'
Warning message: unloaded all rxode2 dlls before loading the current DLL
 *** caught segfault ***
address (nil), cause 'memory not mapped'
Traceback: 1: rxSolveSEXP(...)
=== EXIT STATUS: 139 ===
```

Compiling *without* solving exits 0 -- `rxode2()` and `rxModelVars()` both
"succeed" on the broken model.

```r
# with `cli` absent from .libPaths()
library(rxode2)
```

```
terminate called after throwing an instance of 'Rcpp::LongjumpException'
=== EXIT STATUS: 134 ===
```

### Expected

An R condition naming the missing package, as already happens for the twelve
`Imports:` that *are* in `NAMESPACE` (e.g. omitting `lotri` or `PreciseSums`
gives a clean `there is no package called 'lotri'` at `library()`).

### `rxode2ll` -> SIGSEGV

`src/codegen.c:247-253` emits an `R_GetCCallable()` into every generated model's
`_assignFuns0()` for **every** row of the parse table -- including all the
`rxode2ll` rows -- unconditionally, and in reverse order (hence the last row,
`rxLlikCauchyDscale`, is the symbol reported):

```c
for (int i = Rf_length(_rxode2parse_functionName); i--;) {
  sAppend(&sbOut,"  %s = (%s) R_GetCCallable(\"%s\", \"%s\");\n", ...);
}
```

`src/codegen.h:457` calls `_assignFuns0()` from the model's `R_init_<lib>()`.
`R_GetCCallable` calls `Rf_error()`, so `R_init` longjmps out **before** the
closing `_solveData = _getRxSolve_();`, leaving every global pointer NULL.

The failure is then masked at `R/rxode2.R:2107-2116`:

```r
.tmp <- try(dynLoad(.cDllFile), silent = FALSE)
if (inherits(.tmp, "try-error")) { rxUnloadAll()
  .tmp <- try(dynLoad(.cDllFile), silent = TRUE)   # SUCCEEDS
```

R keeps a DLL in its loaded table after `R_init` errors, so the retry never
re-runs `R_init` and simply reports success. `is.loaded(model_vars)` is true
(`R_init0` registers routines *before* `_assignFuns0`), so the half-initialised
model is accepted and the first `rxSolve()` calls through NULL.

`.nsToLoad()` (`R/utils.R:947`) already runs at exactly the right checkpoint --
`dynLoad()` (`src/rxData.cpp:681`) calls it just before `dyn.load` -- but it
discards its `requireNamespace()` results.

### `cli` -> SIGABRT

`src/rxData.cpp:466-467` has file-scope Rcpp globals:

```cpp
Environment cliNS = loadNamespace("cli");
Function cliAlert0 = as<Function>(cliNS["cli_alert_info"]);
```

Dynamic initialisers run inside `dlopen()`'s `_dl_init`, where the R error
becomes an `Rcpp::LongjumpException` with no handler, reaching `std::terminate`.
gdb backtrace: `__static_initialization_and_destruction_0 () at rxData.cpp:466`
<- `_dl_init` <- `dlopen(".../rxode2/libs/rxode2.so")`.

(`src/qs.cpp` and `src/checkmate.cpp` already avoid this by loading their
namespaces inside a function instead.)

## Which packages are affected

Measured by building one isolated symlink library per package, omitting that
package, then running: parse a ui model, compile+solve an ODE model, solve a ui
model with `add()`, and `rxS()`.

| outcome | packages |
|---|---|
| **SEGFAULT (139)** | `rxode2ll` |
| **SIGABRT (134)** | `cli` |
| clean error at `library()` | `PreciseSums`, `dparser`, `lotri`, `checkmate`, `data.table`, `memoise`, `ggplot2`, `Rcpp`, `RcppParallel`, `sys`, `vctrs`, `backports` |
| that workload unaffected | `qs2`, `inline`, `mirai`, `rex`, `symengine`, `units` |

"Unaffected" means unaffected *by that workload*. `qs2`, `inline` and `mirai` are
reached only via `::` elsewhere (`R/rxRaw.R`, `R/rxSymInv.R:244`, `R/rxOom.R`) and
were **not** exercised -- not proven safe. `rex` is used only at build time
(`R/dsl.R`), so it is genuinely runtime-optional.

Note `rxode2ll` is `Imports:` (`DESCRIPTION:125`) but has **no**
`import()`/`importFrom()` in `NAMESPACE`, and R only enforces the NAMESPACE
imports at load -- so it behaves exactly like an unguarded `Suggests:`. Same for
`cli`. Both are used via `::` in R code, which is presumably why neither ended up
in `NAMESPACE`.

## Environment

```
rxode2   5.1.7 (origin/main @ 7d74e29e4, built from source)
R        4.6.1 (2026-06-24), x86_64-pc-linux-gnu
OS       Ubuntu 24.04.4 LTS
```

## Suggested fix

Make `.nsToLoad()` stop instead of discarding its results, and load the `cli`
namespace lazily. A PR doing both, with a subprocess-based regression test, is
linked below.

Worth considering separately: emitting the `llik*()` lookups only for the
functions a model actually uses, so a plain ODE model does not need `rxode2ll`
at all.

billdenney:
Fixes #1204

Fixes two independent ways a missing `DESCRIPTION` `Imports:` package crashes the
R process instead of erroring: `rxode2ll` missing gives SIGSEGV (139) on the
first solve of *any* compiled model, and `cli` missing gives SIGABRT (134) inside
`library(rxode2)` itself. Neither is catchable -- `tryCatch()` and testthat die
with the process.

## `rxode2ll` -> SIGSEGV

Every generated model resolves the whole parse table with `R_GetCCallable()` from
its `R_init()`, including the `rxode2ll` rows (`src/codegen.c:247`). That errors
out of `R_init` before the model's globals are assigned, leaving them NULL -- but
R keeps the dll in its loaded table, so the retry in `R/rxode2.R:2107` looks like
it succeeded, the half-initialised model is accepted, and the first solve calls
through NULL.

`.nsToLoad()` already ran at exactly the right checkpoint (`dynLoad()` calls it
immediately before `dyn.load`) but threw its `requireNamespace()` results away.
It now stops and names the missing package.

## `cli` -> SIGABRT

A file-scope `Environment cliNS = loadNamespace("cli")` runs as a dynamic
initialiser inside `dlopen()`, where the R error becomes an
`Rcpp::LongjumpException` with no handler and reaches `std::terminate`. Now
loaded on first use -- the same shape `checkmate.cpp` already uses for
`loadCheckmate()`/`qtest()`.

## Test

`tests/testthat/test-missing-package.R` runs the workload in a child `Rscript`
against a symlink library with one package hidden, and asserts the child exits
under R's control rather than on a signal. A subprocess is required because a
SIGSEGV/SIGABRT would kill the testthat process itself. It builds its library
from symlinks and never writes to a real one.

Verified **red on `main`** (6 failures: `139 >= 128` + `caught segfault` for
`rxode2ll`, `134 >= 128` + `terminate called` for `cli`) and green with this
change.

The test asserts "an R condition, never a signal" rather than requiring an error,
because a package reached only lazily need not be wanted by a given workload at
all -- an uneventful run is a legitimate pass.

## Verification

- `test-missing-package.R`: 9 pass on this branch, red on `main`.
- Sweep of `test-par-solve.R`, `test-lincmt-solve.R` (1569), `test-omega-chol.R`
  (435), `test-ui-simulation.R`, `test-solver-basic.R`, `test-sim-zeros.R`,
  `test-reset.R`, `test-et.R`, `test-deSolve-events.R`,
  `test-multisim-shared-event-oob.R`, `test-nopred-ui.R`: no change versus `main`.
  `test-et.R` has 10 failures identically before and after (pre-existing).

**One limitation, stated plainly:** I could not construct a call that reaches
`cliAlert()` -- the R layer strips zero `thetaMat`/`omega`/`sigma` before the C
code tests `_globals.zeroOmega` etc., which is presumably why no test covers
those messages. So the `cli` change is verified by the crash disappearing and by
matching the already-exercised `checkmate.cpp` idiom, not by exercising the alert
itself. Worth a maintainer eye.

## Not included

Emitting the `llik*()` lookups only for functions a model actually uses -- that
would remove the `rxode2ll` dependency for plain ODE models entirely, but it is a
codegen change with a wider blast radius and belongs in its own PR.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,6 +24,16 @@
   `RcppParallel` 6.2.0 restored the TBB library on Windows rather than
   dropping it.)
 
+- A missing `Imports:` package no longer crashes the R process.  Every compiled
+  model resolves `rxode2ll`'s symbols with `R_GetCCallable()` from its
+  `R_init()`, which errors out of `R_init()` and leaves the model's globals
+  `NULL`; R keeps the dll loaded regardless, so the half-initialized model was
+  accepted and the first solve segfaulted.  `rxode2` now stops with a message
+  naming the missing package before loading a model dll.  Separately, a
+  file-scope `loadNamespace("cli")` ran inside `dlopen()`, where the R error
+  raised when `cli` was absent became an unhandled C++ exception and aborted
+  `library(rxode2)` itself; the `cli` namespace is now loaded on first use.
+
 ## Bug fixes
 
 - `rxCompile()` now re-parses the model it is handed whenever the parser's
--- a/R/utils.R
+++ b/R/utils.R
@@ -945,10 +945,21 @@ is.latex <- function() {
 
 
 .nsToLoad <- function() {
-  vapply(rxode2parseGetPackagesToLoad(),
-         function(pkg) {
-           requireNamespace(pkg, quietly = TRUE)
-         }, logical(1))
+  .pkgs <- rxode2parseGetPackagesToLoad()
+  .ok <- vapply(.pkgs,
+                function(pkg) {
+                  requireNamespace(pkg, quietly = TRUE)
+                }, logical(1))
+  # Must be an error: every compiled rxode2 model resolves these packages'
+  # symbols with R_GetCCallable() from its R_init(), which errors out of R_init
+  # and leaves the model's globals NULL.  R keeps the dll loaded anyway, so the
+  # half-initialized model is accepted and the first solve segfaults.
+  if (!all(.ok)) {
+    stop("rxode2 models require package(s) that are not installed: ",
+         paste0("'", .pkgs[!.ok], "'", collapse = ", "),
+         call. = FALSE)
+  }
+  .ok
 }
 
 #' Check if a language object matches a template language object
--- a/src/rxData.cpp
+++ b/src/rxData.cpp
@@ -463,16 +463,29 @@ Function loadNamespace("loadNamespace", R_BaseNamespace);
 
 Function requireNamespace("requireNamespace", R_BaseNamespace);
 
-Environment cliNS = loadNamespace("cli");
-Function cliAlert0 = as<Function>(cliNS["cli_alert_info"]);
+// Loaded on first use, not at file scope: a file-scope initializer runs inside
+// dlopen(), where an R error (as when 'cli' is not installed) becomes an
+// Rcpp::LongjumpException with no handler and calls std::terminate, aborting R
+// during library(rxode2).  Same shape as loadCheckmate() in checkmate.cpp.
+Environment cliNS;
+bool loadedCliNs = false;
+
+static void loadCli() {
+  if (!loadedCliNs) {
+    cliNS = loadNamespace("cli");
+    loadedCliNs = true;
+  }
+}
 
 extern "C" void cliAlert(const char *format, ...) {
   va_list args;
   char buffer[256];
   va_start(args, format);
   vsnprintf(buffer, 256, format, args);
-  cliAlert0(wrap(buffer));
   va_end (args);
+  loadCli();
+  Function cliAlert0 = as<Function>(cliNS["cli_alert_info"]);
+  cliAlert0(wrap(buffer));
 }
 
 extern "C" SEXP _rxode2_rxRmvn0(SEXP A_SEXP, SEXP muSEXP, SEXP sigmaSEXP,
--- a/tests/testthat/test-missing-package.R
+++ b/tests/testthat/test-missing-package.R
@@ -1,0 +1,100 @@
+# A missing DESCRIPTION Imports: package must produce an R condition, never a
+# signal.  These run in a child Rscript because a SIGSEGV/SIGABRT would kill the
+# testthat process itself, so the assertion is on the child's exit status.
+#
+# The two regressions guarded here:
+#   rxode2ll missing -> SIGSEGV (139) on the first solve of any compiled model.
+#     Every generated model resolves rxode2ll's symbols with R_GetCCallable()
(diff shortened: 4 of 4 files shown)


mattfidler:
per discussin

PR #1209, opened 2026-08-07
fix: resolve rxode2ll llik symbols only in models that call one

billdenney:
@mattfidler, this is likely low importance based on our discussion in the other issue about missing packages.

Follow-up to the missing-`Imports:` crash PR, which that PR deliberately left out.
Stacked on `fix/missing-imports-crash`; base this on that branch.

## What was happening

Every generated model resolved the **entire** parse table with `R_GetCCallable()`
from its `R_init()` (`src/codegen.c:237`), including all 39 of rxode2ll's
`llik*()` rows, whether or not the model used any of them. So a plain ODE with no
`llik*()` call anywhere still depended on rxode2ll -- and not softly:
`R_GetCCallable` calls `Rf_error()`, which longjmps out of `R_init` before the
model's globals are assigned, while R keeps the dll in its loaded table.

## The change

rxode2ll contributes only `llik*()` functions, and the parser already records
whether a model calls one: `handleLlFunctions()` (`src/parseFuns.h:385`) sets
`tb.nLlik`, which is reset per parse (`src/tran.c:460`) and exported in the model
vars. So the guard is a single condition on an existing, already-trusted signal --
no new tracking:

```c
if (tb.nLlik == 0 &&
    !strcmp(R_CHAR(STRING_ELT(_rxode2parse_functionPackageName, i)), "rxode2ll")) {
  continue;
}
```

`handleLlFunctions()` runs inside the recognised-function branch of
`handleBadFunctions()`, so every `llik*`/`llikX*` call in the parsed model text
sets it -- including models rxode2 generates itself (error models, focei), since
those are re-parsed as text. `_rxode2parse_packages` is only `"rxode2"`, so these
lookups were the sole rxode2ll reference in generated code.

## Evidence

Generated C, counting `rxode2ll` references:

| model | before | after |
|---|---|---|
| `d/dt(x) <- -0.5*x;` | 39 | **0** |
| `ll <- llikNorm(0, 1, 2); d/dt(x) <- -0.5*x;` | 39 | 39 |

With rxode2ll genuinely absent and dropped from the `packagesToLoad` guard, a
plain ODE model:

- on `main`: `*** caught segfault ***`, **exit 139**
- with this change: `SOLVED: 0.6065312` (= e^-0.5, correct), **exit 0**

`llikNorm()` returns an identical value before and after (`-1.737086`).

## Tests

Added to `test-llik.R`, next to the existing `nLlik` flag test. It asserts the
plain model generates exactly zero rxode2ll references and the `llik` model
generates one per rxode2ll row -- the expected count is derived from
`rxode2parseGetTranslation()` rather than hard-coded, so it stays exact if the
table grows. **Red on `main`** (`Expected .llRefs(.plain) to equal 0L`).

Sweep, all green, 0 failures: `test-llik.R` (388), `test-lincmt-solve.R` (1569),
`test-omega-chol.R` (435), `test-par-solve.R` (357), `test-focei-variances.R`,
`test-ui-focei-ord.R`, `test-ui-pred-err.R`, `test-ui-simulation.R`,
`test-solver-basic.R`, `test-missing-package.R`, `test-sim-zeros.R`,
`test-reset.R`, `test-rm-dll.R`, `test-ind-alloc.R`, `test-dollar-names.R`.

## Scope, stated plainly

`.nsToLoad()` still requires rxode2ll for every model load, so **this does not by
itself change what a user sees when rxode2ll is missing** -- that remains the
clean error from the parent PR. What it changes is that the generated code no
longer references rxode2ll for non-llik models, which is what makes the guard
narrowable later, and it drops 39 symbol lookups from every non-llik model's load.

Making the guard itself model-aware would be the natural next step, but the model
vars that carry `nLlik` are only read *after* `dynLoad()`, so it needs a
restructure rather than a condition -- left alone here.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -34,6 +34,11 @@
   raised when `cli` was absent became an unhandled C++ exception and aborted
   `library(rxode2)` itself; the `cli` namespace is now loaded on first use.
 
+- Generated models only resolve `rxode2ll`'s `llik*()` symbols when the model
+  actually calls one, instead of resolving all of them in every model.  A model
+  with no `llik*()` no longer depends on `rxode2ll` at all, and skips that many
+  `R_GetCCallable()` lookups when it loads.
+
 ## Bug fixes
 
 - `rxCompile()` now re-parses the model it is handed whenever the parser's
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,6 +24,16 @@
   `RcppParallel` 6.2.0 restored the TBB library on Windows rather than
   dropping it.)
 
+- A missing `Imports:` package no longer crashes the R process.  Every compiled
+  model resolves `rxode2ll`'s symbols with `R_GetCCallable()` from its
+  `R_init()`, which errors out of `R_init()` and leaves the model's globals
+  `NULL`; R keeps the dll loaded regardless, so the half-initialized model was
+  accepted and the first solve segfaulted.  `rxode2` now stops with a message
+  naming the missing package before loading a model dll.  Separately, a
+  file-scope `loadNamespace("cli")` ran inside `dlopen()`, where the R error
+  raised when `cli` was absent became an unhandled C++ exception and aborted
+  `library(rxode2)` itself; the `cli` namespace is now loaded on first use.
+
 ## Bug fixes
 
 - `rxCompile()` now re-parses the model it is handed whenever the parser's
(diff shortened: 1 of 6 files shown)


mattfidler:
Still not sure this complexity is worth it...?

billdenney:
Fair that it doesn't add much

PR #1217, opened 2026-08-10
fix(eventSens): translate THETA[n]/ETA[n] in every emitted C line (#1196)

### LINKED ISSUE #1196: eventSens: .rxEventSensCExpr() only translates THETA[n]/ETA[n] from the first vector element, emitting uncompilable C
Reported downstream as nlmixr2/nlmixr2#408 ("Error building model - food effect not running?").

## Summary

`.rxEventSensCExpr()` (`R/eventSens.R`) only collects `THETA[n]`/`ETA[n]` tokens from the **first** element of its (vectorized) `expr` argument. Any index that appears solely in a later element is never rewritten to the codegen local `_THETA_n_`/`_ETA_n_`, so raw symengine array syntax leaks into the generated C and the model fails to compile:

```
error: 'ETA' undeclared (first use in this function)
error: 'THETA' undeclared (first use in this function)
```

This is a real codegen bug — it reproduces on Linux with a perfectly good toolchain. The reporter in nlmixr2#408 was sent down an Rtools rabbit hole because of the diagnostic wording (filed separately).

## Root cause

`R/eventSens.R:806`:

```r
.toks <- unique(regmatches(expr, gregexpr(paste0(kind, "\\[[0-9]+\\]"), expr))[[1]])
```

`expr` is a character vector — one entry per event-sensitivity assignment line — but `[[1]]` keeps only the matches from element 1. `gsub()` then substitutes those tokens across all elements, so whatever happens to appear first is translated and the rest silently survive as `ETA[n]`/`THETA[n]`.

Minimal demonstration against the installed function:

```r
e <- c("exp(ETA[4]+THETA[4])*(FOOD!=0)", "exp(ETA[5]+THETA[5])*(FOOD==0)")
rxode2:::.rxEventSensCExpr(e, character(0))
#> [1] "exp(_ETA_4_+_THETA_4_)*(FOOD!=0)"  "exp(ETA[5]+THETA[5])*(FOOD==0)"
#>                                              ^^^^^^^^^^^^^^^^ not translated
```

Resulting generated C (from `rxode2::rxLastCompile()$c`):

```c
void ..._dDur(int _cSub, double __t, double *__zzStateVar__, double *_dDurSave){
  _dDurSave[3] = exp(_ETA_4_+_THETA_4_)*(FOOD!=0);   // OK
  _dDurSave[4] = exp(ETA[5]+THETA[5])*(FOOD==0);     // does not compile
}
```

## Suggested fix

Collect tokens across every element:

```r
.toks <- unique(unlist(regmatches(expr, gregexpr(paste0(kind, "\\[[0-9]+\\]"), expr))))
```

Verified this produces the correct output for the vector above:

```
[1] "exp(_ETA_4_+_THETA_4_)*(FOOD!=0)" "exp(_ETA_5_+_THETA_5_)*(FOOD==0)"
```

## Reproducible example

Needs a model where a **dosing event modifier** (`dur()`, `f()`, ...) depends on two or more distinct etas, so that `.rxEventSensCLines()` emits more than one assignment line for the same buffer:

```r
library(nlmixr2)

d <- nlmixr2data::theo_sd
d$FOOD <- ifelse(d$ID %% 2 == 0, 1, 0)
d$RATE <- ifelse(d$EVID == 1, -2, 0)   # duration dosing

mod <- function() {
  ini({
    tka <- 0.5; tcl <- 1; tv <- 3.5; tD1a <- 0.1; tD1b <- 0.1
    eta.ka ~ 0.1; eta.cl ~ 0.1; eta.v ~ 0.1; eta.D1a ~ 0.1; eta.D1b ~ 0.1
    prop.sd <- 0.1
  })
  model({
    is_fed <- (FOOD != 0); is_fast <- (FOOD == 0)
    ka <- exp(tka + eta.ka); cl <- exp(tcl + eta.cl); v <- exp(tv + eta.v)
    D1 <- is_fed * (exp(tD1a + eta.D1a)) + is_fast * (exp(tD1b + eta.D1b))
    dur(depot) <- D1
    d/dt(depot)  = -ka * depot
    d/dt(center) =  ka * depot - cl/v * center
    cp <- center/v
    cp ~ prop(prop.sd)
  })
}

nlmixr2(mod, d, est = "posthoc")   # Error: error building model
```

## Scope

Narrowed by variant testing:

| variant | result |
| --- | --- |
| two-eta expression in `dur()` | **fails** |
| two-eta expression in `f()` | **fails** |
| same expression used only in the ODE / `cp` (no event modifier) | OK |
| single-eta `dur()` / `f()` / `lag()` | OK |

So it requires at least two distinct etas contributing separate lines to the same event-sensitivity buffer. The `some etas defaulted to non-mu referenced` warning that accompanies these models is a red herring — mu-referencing both branches on simple lines yields byte-identical C and fails the same way.

## Workaround (for reference)

Restructuring so every emitted line contains all the indices avoids it, e.g. folding the branches inside a single `exp()`:

```r
D1 <- exp(is_fed * (tD1a + eta.D1a) + is_fast * (tD1b + eta.D1b))
```

This is algebraically identical when the indicators are mutually exclusive, and it compiles and fits.

## Environment

- R 4.6.1, Linux, gcc 14.2.0 (working toolchain)
- rxode2 5.1.7, nlmixr2est 7.0.2, nlmixr2 6.0.0
- Reporter in nlmixr2#408: Windows, R 4.6.1, rxode2 5.1.6, nlmixr2est 7.0.2

## Regression test suggestion

A test asserting that generated event-sensitivity C contains no bare `ETA[`/`THETA[` would cover this directly, plus a unit test on `.rxEventSensCExpr()` with a length-2 input.

mattfidler:
Fixes #1196.

## The bug

`.rxEventSensCExpr()` (`R/eventSens.R`) rewrites nlmixr2's indexed
`THETA[n]`/`ETA[n]` parameters into the codegen locals `_THETA_n_`/`_ETA_n_`.
Its `expr` argument is a character **vector** -- one entry per emitted C
assignment line -- but the token collection subset the `gregexpr()` match list
with `[[1]]`, so only the indices used by the *first* element were gathered.

A dosing modifier (`dur()`, `f()`, `alag()`, ...) that depends on two or more
estimated parameters emits one assignment line per parameter into the same
buffer, so any index appearing only in a later line survived as raw symengine
array syntax and the generated C did not compile:

```c
_dDurSave[0] = exp(_ETA_4_+_THETA_4_)*(FOOD!=0);   // OK
_dDurSave[1] = exp(ETA[5]+THETA[5])*(FOOD==0);     // 'ETA' undeclared
```

This is what the food-effect model in nlmixr2/nlmixr2#408 hit ("error building
model"); the reporter was sent down an Rtools rabbit hole, but it reproduces on
a perfectly good Linux toolchain.

## The fix

Collect the tokens across every element:

```r
.toks <- unique(unlist(regmatches(expr, gregexpr(paste0(kind, "\\[[0-9]+\\]"), expr))))
```

All five `.rxEventSensCLines()` helpers pass a vector through this one function,
so the single change covers the 1st-, 2nd-, and 3rd-order buffers as well.

## Tests

`tests/testthat/test-event-sensitivities.R`:

- a unit test on `.rxEventSensCExpr()` with a length-2 input (plus `plainParams`,
  zero-length, and no-token inputs);
- `dur()`/`f()`/`alag()` models whose modifier is built from two etas: the
  generated C must contain no bare `ETA[`/`THETA[`, and must reference
  `_ETA_5_`/`_THETA_5_`;
- a finite-difference check that the line which used to be left untranslated
  carries the *right* value -- with `FOOD=0` the duration is
  `exp(ETA[5]+THETA[5])`, so the jump sensitivity wrt `ETA[5]` matches central
  differences (and wrt `ETA[4]` is zero).

All three fail on `main` (the models do not compile).

Also moves a misplaced `@noRd` roxygen block: it documented
`.rxEventSensCLines()` but sat above `.rxEventSensCExpr()`, which now has its
own.

## Verification

- `testthat::test_local(filter='sens')`: 365 pass, 0 fail
- `testthat::test_local(filter='focei|jump|codegen|lincmt')`: 6326 pass, 0 fail
- `testthat::test_local(filter='event-sensitivities|adjoint|dde')`: 3142 pass, 0 fail
- Two independent Antigravity (Gemini) review rounds over the diff: no findings.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -152,6 +152,16 @@
   restored SAEM fit failing with "The following parameter(s) are required for
   solving: eta.v, eta.cl".
 
+- Event ("jump") sensitivities now compile when a dosing modifier (`dur()`,
+  `f()`, `alag()`, ...) depends on more than one estimated parameter.  Each such
+  parameter contributes its own assignment line to the same generated buffer,
+  but the rewrite of nlmixr2's indexed `THETA[n]`/`ETA[n]` to the codegen locals
+  `_THETA_n_`/`_ETA_n_` only collected the indices used by the *first* line, so
+  an index appearing only in a later line survived as raw symengine array syntax
+  and the model failed with "'ETA' undeclared".  This hit any model with, say, a
+  food-effect duration built from two etas, whether or not the parameters were
+  mu-referenced (#1196).
+
 ### Delay differential equations
 
 - `rxOptExpr()` no longer fails on a `past(state, tau)` whose delay duration is
--- a/R/eventSens.R
+++ b/R/eventSens.R
@@ -787,18 +787,16 @@
        durQ = .buildQ(map$durCmt, "dur", .q2All))
 }
 
-#' Generate the C assignment lines for the dLag / dF functions
+#' Rewrite indexed nlmixr2 parameters into their codegen locals
 #'
-#' Produces the body assignment lines writing each dosing-parameter total
-#' derivative into a flat per-subject scratch buffer indexed
-#' `(cmt0 * nSensParam + paramIdx)`.  The lines are inserted into the generated
-#' C verbatim; the only rewrite needed is nlmixr2's indexed
-#' `THETA[n]`/`ETA[n]` parameters, which `.rxEventSensCExpr()` maps to the
-#' codegen locals `_THETA_n_`/`_ETA_n_`.
+#' Maps nlmixr2's indexed `THETA[n]`/`ETA[n]` to the codegen locals
+#' `_THETA_n_`/`_ETA_n_`, or to the plain `THETA_n_`/`ETA_n_` when the model
+#' declares that name itself.
 #'
-#' @param info An `.rxEventSensInfo()` result (mode + map + derivs).
-#' @return list with `nSensParam`, `paramIdx` (named 0-based), and character
-#'   vectors `lag` and `f` of C assignment lines; `NULL` if `info` is `NULL`.
+#' @param expr Character vector of C expressions, one per assignment line.
+#' @param plainParams Parameter names the model declares plainly (no leading
+#'   underscore).
+#' @return `expr` with every indexed parameter rewritten.
 #' @noRd
 .rxEventSensCExpr <- function(expr, plainParams = character(0)) {
   # THETA[n]/ETA[n] map to the codegen locals _THETA_n_/_ETA_n_ unless the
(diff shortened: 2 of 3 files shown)


mattfidler:
Seems a bit hackish but it works.

PR #1228, opened 2026-08-11
feat(api): expose per-compartment InfusionRate and on/off accessors

mattfidler:
## Why

The generated `dydt()` reads `_IR[]` straight into the right-hand side and gates every line on `_ON[]`:

```c
__DDtStateVar__[__DDT0__] = ((double)(_ON[__DDT0__]))*(_IR[__DDT0__] - cl/_div0(v)*centr);
```

Neither was reachable through the C function-pointer table. A downstream package that evaluates `dydt()` or `calc_jac()` **outside** an integration — over a fixed grid of states rather than along a solve — therefore had no way to set the dosing state, and silently got a right-hand side with no dosing at all.

This came up building a `magi` (manifold-constrained Gaussian process ODE inference) estimation method in babelmixr2, which needs `f`, `∂f/∂x` and `∂f/∂θ` evaluated at arbitrary states on a discretization grid rather than along a trajectory.

## What

Adds four accessors:

| | |
|---|---|
| `getIndInfusionRate(ind, i)` / `setIndInfusionRate(ind, i, val)` | compartment `i`'s infusion rate |
| `getIndOn(ind, i)` / `setIndOn(ind, i, val)` | whether compartment `i` is on |

Both are bounds-checked against `op->neq + op->extraCmt` — the size both buffers are allocated at in `_setIndPointersByThread()` — and guarded against the `NULL` both are set to when that count is zero.

## Compatibility

The four slots are **appended** (`nVec` 92 → 96). A package built against an older `rxode2ptr.h` stops reading at slot 91 and is unaffected; the existing name-freeze contract is preserved.

Covered by the existing invariants in `tests/testthat/test-event-sensitivities-api.R` (every slot filled/named/unique, every slot read exactly once by `iniRxodePtrs0()` with a matching name stem, existing names unchanged), plus an explicit presence assertion for the new names.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,16 @@
 
 ## New features
 
+- The C function-pointer table (`.rxode2ptrs()`) gained
+  `getIndInfusionRate()`/`setIndInfusionRate()` and `getIndOn()`/`setIndOn()`,
+  which read and write a compartment's infusion rate and on/off flag.  The
+  generated `dydt()` reads both -- `_IR[]` enters the right-hand side directly
+  and `_ON[]` gates every line -- so a downstream package evaluating `dydt()` or
+  `calc_jac()` *outside* an integration (for example over a fixed grid of states
+  rather than along a solve) previously had no way to set them, and silently got
+  a right-hand side with no dosing.  The slots are appended, so packages built
+  against an older header are unaffected.
+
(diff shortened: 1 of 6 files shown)


mattfidler:
nope.

PR #1281, opened 2026-08-23
linCmt(): hybrid sensitivity strategy (rxSolve(linCmtSensStrategy=)) + partial-mask steady-state fix

mattfidler:
Stacks on #1280 (`lincmt-auto-sens-count`); retarget to `main` after #1280 merges. Phase 4.3 of the linCmt() speed plan, revived after the hybrid prototype was re-measured with realistic requested-direction counts.

## What it does

`rxSolve(linCmtSensStrategy = c("auto", "sequential", "hybrid"))`. A subject's rows up to the start of its trailing run of observations are solved by the existing sequential kernel (whatever `linCmtSensType` picks); from that row on the per-row amount+Jacobian buffer is filled by superposition over the phase-1 end state, one reverse-mode nest per row that sweeps only the pred (central) output -- one adjoint sweep per observation instead of one forward pass per requested direction or one sweep per compartment.

Design (all in `src/linCmt.cpp`):
- **Pre-pass**: found lazily on the subject's first computed row (`ind->ix` is not sorted at `iniSubject()` time): the trailing observation-only run (`isObs(evid)`) is phase 2. Engaging is a performance choice only; rows the scan did not see (pushed doses, modeled times) are handled as ordinary rows -- never a mid-solve bail-out.
- **Phase 2 filler** (`linCmtHybRow`): the 4.2 superposition design -- the list primes from the buffer's own amounts + the sequential kernel's Jacobian at the boundary (`A0 + J0*(theta - theta0)`, exact since the kernel is linear in its initial state), an amount added since the last fill is a new term, a change in the live rate vector a rate-step term, a rate decrease or the `linCmtHybridMaxActive` ceiling collapses the list into one primed term.
- **Buffer contract**: amounts for every compartment are exact; the Jacobian block holds the central row only (other rows written as 0). Nothing downstream reads them: `d(cp)/d(theta)` uses the central row, the carry sentinels compute their transition matrix from theta, and any row the sequential kernel takes while the phase-2 state is live (steady state, modeled-lag/extra dose, a theta change between rows) is preceded by `linCmtHybFlush()`, which rewrites the previous row's block with the full Jacobian. A model that reads raw rows (`linCmtB(which1 >= 0, which2 >= 0)`, flagged by the parser as model-variables flag 16 `linCmtBraw`) gets every row swept.
- **Multi-call rows**: generated code calls `linCmtB(-1, -1)` several times per row; the list is re-evaluated, never re-primed. The per-thread state is dropped at every subject's pass start (an earlier solve's `(id, idx, tprior)` can coincide).
- **Engage rule** (`auto`): pure `linCmt()` solve, `ndiff != 0`, AD sensType, not a steady-state/extra-dose row, >= 2 compartments, trailing run >= `linCmtHybridMinObs` (2), requested directions >= `linCmtHybridMinDirs` (3, measured: two directions are parity). `"hybrid"` forces it, `"sequential"` is the old behavior. A subject with a pending steady-state-infusion turn-off or modeled lag stays sequential. `linCmtHybStats()` exposes subjects/rows/doses/rateSteps/consolidations/flushes/fullRows.

Also fixes a pre-existing bug found by the validation: a model reading only some `linCmtB()` directions (a FOCEi inner model with fewer etas than parameters) solved every row as `NA` after a steady-state dose -- `updateJfromJs()` left the unrequested columns at the slot's initial `NA`, which `saveJac()` carried into the next row's Alast reconstruction (`src/linCmt.h`, first commit, `test-lincmt-partial-mask-ss.R`: 6 configs x 3 masks x AD/ADr x ss=1 infusion/bolus + ss=2).

## Validation (real `rxSolve()` path, `bench/lincmt_hybrid_production.R` + `tests/testthat/test-lincmt-sens-hybrid.R`)

hybrid vs sequential to round-off (< 1e-9, typically 1e-13..1e-15) on: 6 configs x {bolus, infusion, mixed} x k = 1,2,3 requested directions (non-uniform spacing, 3 subjects); steady-state regimens (ss=1 infusion/bolus, ss=2); observations interleaved with doses (phase 2 = trailing run only, counted); a model reading raw Jacobian rows (every row swept, counted); a generated-style model with several value calls per row; state isolation across solves; 24 subjects at 11 threads bit-identical to 1 thread over repeated runs. Counters assert the hybrid actually took the rows. All 18 `test-lincmt*.R` files pass (6,635 assertions); carry benches unchanged (phase2 3.078e-11, phase3b2 11/11).

## Measured speedup (44 subjects, doses x observations grid, sequential under #1280's auto rule vs hybrid)

| config | k | 1 thread | 11 threads |
|---|---|---|---|
| 2-cmt oral | 2 | 1.02-1.11x | 0.96-1.21x |
| 2-cmt oral | 3 | 1.03-1.22x | 1.02-1.25x |
| 3-cmt oral | 2 | 0.98-1.02x | 0.75-1.06x |
| 3-cmt oral | 3 | 1.06-1.32x | 1.02-1.34x |

Honest reading: the win is 1.1-1.3x with three requested directions on the dose-heavy-then-rich-sampling shape (e.g. 3-cmt oral, 20 doses / 400 obs: 1.215s -> 0.921s single-thread), parity with two, below the prototype's 1.3-1.5x estimate -- the per-observation reverse nest carries allocation overhead the fvar passes do not. Hence `linCmtHybridMinDirs = 3` by default, so the `auto` default never engages where it does not pay.

## Not engaged / deferred

- Subjects with a pending steady-state-infusion turn-off or a modeled lag stay sequential (exact, just no gain); refining the pending-dose test to "nothing pending after this row" would recover them.
- Phase 2 is the trailing observation run only; doses continuing through the sampling window keep those rows sequential. Extending the boundary earlier (doses as phase-2 terms with consolidation) is possible with the same machinery but unmeasured.
- The phase-2 nest could be made allocation-free (pre-sized var buffers) -- the likely route to the prototype's estimate.

## Update: allocation-free observation-phase kernel (commits 09c313859, 207ad3cce, 9421fb08b)

**Profile (callgrind, per phase-2 row, before):** 164k instructions per row, 84% inside the reverse-mode nest: `linCmtStan2<var>` tape build 52%, `computeSolComp2<var>` 11%, `set_zero_all_adjoints` 7%, vari allocation / `vector<vari*>` push 15-20%, `__tls_get_addr` (the per-thread tape under `STAN_THREADS`) 5%. The cost was the tape build, not heap allocation of Eigen temporaries.

**What changed:** the reverse nest is gone. The theta-only constants of the closed form (k10, or the 2/3-cmt eigen-decomposition L/C, and ka) and their derivatives are computed once per subject window by `npars` forward passes (`linCmtHybWinFill`); each observation row then runs one `fvar` pass per *requested* direction through the dt-dependent tail of the kernel (`linCmtStanNTail`, split out of `linCmtStanN` with the sequential kernels unchanged), giving every compartment's Jacobian row -- no tape, no allocation (`restoreJacTo()` reuses a per-thread matrix). A hybrid subject's dose phase is always rolled through in forward mode. `linCmtHybStats()` gained a `windows` counter (one fill per subject window).

**Build flags matter:** `devtools::load_all()` builds with `-O0` (pkgbuild's debug flags override `-O3`), so every earlier timing on this branch was of unoptimized code. The bench now loads with `compile = FALSE` after `pkgbuild::compile_dll(debug = FALSE)`. On the optimized build the same 2-cmt-oral k=3 100x200 cell went from 0.443 s (sequential) / 0.387 s (hybrid) to 0.071 / 0.025 s.

**Optimized build, 44 subjects, dose-heavy x obs-rich grid** (ratio ranges over the 9 cells; `auto` = sequential under PR #1280's count rule, `forward` = sequential forced to `"AD"`):

| config | k | threads | old hybrid/auto (-O0) | hybrid/auto | hybrid/forward | auto/forward |
|---|---|---|---|---|---|---|
| 2cmt-oral | 2 | 1 | 1.02-1.11 | 1.00-1.10 | 1.00-1.17 | 0.86-1.00 |
| 2cmt-oral | 2 | 11 | 0.96-1.21 | 0.74-1.11 | 0.71-1.11 | 1.00-1.10 |
| 2cmt-oral | 3 | 1 | 1.03-1.22 | 2.25-3.33 | 1.00-1.12 | 2.25-3.00 |
| 2cmt-oral | 3 | 11 | 1.02-1.25 | 1.33-2.15 | 0.91-1.06 | 1.33-2.15 |
| 3cmt-oral | 2 | 1 | 0.98-1.02 | 0.90-1.38 | 0.97-1.26 | 0.82-1.24 |
| 3cmt-oral | 2 | 11 | 0.75-1.06 | 0.78-1.20 | 0.78-1.12 | 0.89-1.20 |
| 3cmt-oral | 3 | 1 | 1.06-1.32 | 0.89-1.52 | 0.89-1.33 | 0.95-1.21 |
| 3cmt-oral | 3 | 11 | 1.02-1.34 | 0.96-1.36 | 0.96-1.36 | 0.87-1.14 |

Honest reading: on an optimized build the forward-mode sequential kernel is already within ~1.3x of the value-only solve (the per-row solver overhead is most of the time), so the hybrid's gain over *forward* is modest (1.0-1.3x single-threaded, parity to slightly worse threaded at the small cells). The large 2cmt-oral k=3 ratios against `auto` are because #1280's rule picks reverse mode there, and on this build reverse is 2-3x slower than forward (`auto/forward` column) -- that rule was measured on `-O0` code and should be re-measured on an optimized build. `linCmtHybridMinDirs` now defaults to 2 (hybrid never loses single-threaded at two directions).

Validation unchanged: all 18 `test-lincmt*.R` pass on the optimized build (6,635 assertions; hybrid vs sequential < 1e-9 across configs, regimens, k, SS, 11 threads bit-identical), carry benches pass.

## Update (2026-08-23): full optimized-build sweep, engage rule confirmed

`bench/lincmt_auto_optimized.R` (commits 136213238, 4a3356580) times sequential forward, sequential reverse and the hybrid on an optimized build (`pkgbuild::compile_dll(debug=FALSE)`, -O3; the earlier numbers in this PR and in #1278/#1280 were taken through `devtools::load_all()`, which compiles at -O0), pinned single thread on an idle core, over every kernel (1/2/3-cmt x iv/oral), every parameterization (trans 1/2/11/10, plus 3/4/5 for 2-cmt), requested directions k = 0..npars plus the `cl,v` and `cl,v,ka` masks, and seven event shapes (1 dose/20 obs; 2/200; 100/50; 100/200; 200/400; 10 infusions/100 obs; ss=1 + 5 boluses/100 obs), 44 subjects, non-uniform spacing, 3 reps -- 1386 cells. Every trans reproduces trans 1 to 7e-14; hybrid and reverse match forward to 3e-8.

Hybrid gain vs the best sequential method (forward, see #1280), median [min, max] of fwd/hyb:

| config | heavyDense (100/200) | fewDense (2/200) | small (1/20) | ss | infusion |
|---|---|---|---|---|---|
| 1-cmt iv / oral | 1.02-1.03 | 1.02 | 1.00 | 1.00 | 1.00 |
| 2-cmt iv / oral | 1.02 | 1.03-1.04 | 1.00 | 1.00 | 1.00 |
| 3-cmt iv | 1.09 [1.02, 1.19] | 1.12 [1.06, 1.27] | 1.11 | 1.04 | 1.00 |
| 3-cmt oral | 1.09 [0.98, 1.19] | 1.13 | 1.11 | 1.01 | 1.03 |

By requested directions on the heavy shapes (median gain): 3-cmt oral 1.03 (k=1), 1.06 (k=2), 1.08 (k=3), 1.09, 1.11, 1.13, 1.17 (k=7); 3-cmt iv 1.04 .. 1.15; 1/2-cmt 1.00-1.04 at every k. Trans does not change the picture (3-cmt gains 1.06-1.09 for every trans).

Engage-rule candidates scored by cell loss against the better of {forward, hybrid} per cell: the shipped rule (m >= 2, k >= 2, trailing run >= 2) loses >5% in 49/1386 cells (max 0.77 in one cell that re-timed at a 1.11x gain with 5 reps -- a transient; mean loss 0.7%) and wins >10% in 157 cells; restricting to m >= 3 (87 wrong) or to 3-cmt only (130-202 wrong) is worse because it forfeits the small 2-cmt gains, and never engaging loses >5% in 368 cells. The defaults therefore stay `linCmtHybridMinObs = 2`, `linCmtHybridMinDirs = 2`; a minimum dose count is not supported by the data (the 1-dose/20-obs and 2-dose/200-obs shapes gain as much as the dose-heavy ones on 3-cmt). Threaded cells were not timed in this sweep (another job held the machine); they are recorded as NA and will be filled separately.

Note: this branch still carries #1280's earlier count-based `auto` rule in `src/linCmtSensType.h`; #1280 now resolves `auto` to forward, and this branch needs #1280 merged in after it lands (the hybrid's dose phase already forces forward mode, so its own numbers are unaffected).


mattfidler:
Note: #1282 now removes the strategy machinery this PR introduced (rxSolve(linCmtSensStrategy=), the linCmtHybrid* thresholds and linCmtHybStats(); measured to win nowhere once the sequential path was amortized), while keeping the window/tail kernels and the partial-mask steady-state fix it also contains -- you may prefer to merge the stack together or collapse it rather than merging this PR's strategy surface only to delete it one PR later.

PR #1282, opened 2026-08-23
linCmtB(which1 = -8): carry fast-path pin for event-modifier jumps; amount-read parser registration

mattfidler:
Stacks on #1281 (`lincmt-hybrid`); retarget to `main` after it merges.  Companion to the nlmixr2est carry extension on `lincmt-speed` (#1006), which adds f()/alag() jump channels and every `trans` to the linCmt() sensitivity carry.

## Changes

- `linCmtB(which1 = -8)`: pins the subject's carry pass to the full `-5` transition advance (`ind->linCmtCarryVarying = 2`, the same pin `linCmtCarryLiveTest()` uses) and returns 0.  An event-modifier jump fed to the `-7` add does not telescope the way the tracker-column convention does, so the constant-theta fast path (3b.4) must not skip the M products that propagate it.
- `linCmtCarrySentinelMax()`: capability export nlmixr2est gates the jump codegen on (returns 8).
- Parser: a `linCmtB(which1 = cmt, which2 = -2)` amount read (the carry's `D_r = A_r - M_r A_{r-1}` input) registers no derivative column instead of printing `Unknown linCmtB which2=-2` for every such read.

## Tests

`test-lincmt-carry-sentinel-guards.R` gains a `-8` test (with the pin, `linCmtCarryFastStats()` shows no skipped advance on a constant-theta solve; without it, every advance after the first is skipped); 16/16.  Built and tested optimized (`pkgbuild::compile_dll(debug = FALSE)`).

## Update: sensitivity-amortization project (Phase A + B)

Profiling a dense 2-cmt sensitivity solve (perf + an LD_PRELOAD malloc
counter, optimized build, pinned) found the per-row cost was dominated by
heap allocation, not by the closed-form math: ~235 allocations per row
(sensThetaElt taking theta by value once per parameter per linCmtB call --
a generated inner model makes 11 linCmtB calls per row -- by-value
restore/save paths, and per-row work buffers), with the hoistable
theta-only work (macros2micros + the eigen-decomposition) only ~15% of
kernel CPU.

- **Phase A (`ccd165d29`)**: allocation removal -- Eigen::Ref/const-ref
  signatures, in-place restoreJacTo/restoreFxTo on the restore paths
  (linCmtB and linCmtA), void getJacCp, persistent forward-mode work
  buffers. 235 -> 40 allocations/row.
- **Phase B (`ceea5e59b`)**: `linCmtSeqTailJac()` -- every ordinary
  (non-SS) sequential row now evaluates from the theta-keyed window of
  hoisted constants (shared with the hybrid strategy) plus one
  dt-dependent tail pass per requested direction; no macros2micros or
  eigen-decomposition per row. Falls back to the full evaluator for SS
  rows, FD families and reverse mode. 40 -> 30 allocations/row;
  `linCmtHybStats()` gains seqTailRows/seqFullRows.

Validation: matches the independently-validated hybrid evaluator to
2.5e-12 over all sens/pred columns; all 18 test-lincmt files pass (6,637
assertions); both carry benches unchanged (3.078e-11; 11/11); nlmixr2est
carry suite green; 1-vs-11-thread solves bit-identical over repeated runs.

Timing caveat: wall-clock A/B was measured under background contention
(load 1.9-3.5 across runs; recorded in the bench provenance), so the
deterministic allocation counts are the primary evidence; a quiet-machine
A/B of `ccd165d29~1` vs HEAD is the outstanding measurement. The profile
now shows the remaining per-row cost concentrated in the per-call
machinery multiplied by the generated model's 11 linCmtB calls per row --
an nlmixr2est codegen consideration outside this PR.

### Quiet-machine A/B (final numbers)

Pinned single core, optimized builds, idle machine (load recorded per
cell; `bench/results/seq_amortize_ab.rds`). Sequential sensitivity path,
us/obs at 400 subjects x 1000 observations:

| config | baseline | +alloc removal | +window (HEAD) | total |
|---|---|---|---|---|
| 2-cmt oral | 10.88 | 8.21 | 6.98 | 1.56x |
| 3-cmt oral | 18.37 | 14.57 | 8.70 | 2.11x |

Value-only floors are nearly unchanged (2.57 -> 2.40 / 3.49 -> 3.40
us/obs), and on the amortized HEAD the hybrid strategy no longer beats
the sequential path (parity at 2-cmt, slower at 3-cmt) -- its remaining
role is a follow-up decision. The 40x100 FOCEi fit cell improves
42.2 -> 29.3 s (1.44x; identical objective) vs the ODE arm's 14.3 s:
the dense-design inversion narrows from 2.9x to 2.0x but stands. The
remaining gap to the ODE sens solve (~0.75 us/obs at this size) is
dominated by the per-call machinery times the generated model's 11
linCmtB calls per row -- closing it is an nlmixr2est codegen question
(deduplicating the repeated value/sens calls), not further kernel work.

## Update: hybrid strategy removed (a9593ca0b..02ce5d2f4)

Once the sequential path itself was amortized, the per-subject hybrid
strategy won nowhere (parity at two compartments, slower at three --
bench/results/seq_amortize_ab.rds), so `rxSolve(linCmtSensStrategy=)`,
the `linCmtHybrid*` thresholds, the phase-2 term-list machinery, the
per-subject pre-pass fields and `linCmtHybStats()` are removed (all were
introduced on this development stack and never released). Kept, because
the amortized sequential evaluator depends on them: the theta-keyed
window (`linCmtWin`/`linCmtWinFill`), the `linCmtStan{1,2,3}Tail`
kernels, `linCmtSeqTailJac` (counters now via `linCmtSeqStats()`), the
partial-mask steady-state fix and the carry sentinels. The serialize
format moves to 5 (format 4's four strategy ints are read and
discarded; `linCmtBraw` stays as parse metadata -- the sequential
evaluator writes the full Jacobian block either way). The hybrid tests
are salvaged into `test-lincmt-seq-tail.R`, anchored against reverse
mode with mechanism counters; the strategy-dependent bench scripts are
marked historical with their commit range.

## Update: last-row value memo (dedup P1-P4)

Counter instrumentation showed the generated inner model executes the
`linCmtB()` value call ~15x per row (5 full computes + 10 restores), not
once -- each compute repeating the full dt Jacobian.  A per-thread
last-row memo (keyed on subject, row, time, pass, steady-state/H state
and the raw arguments; invalidated by sentinels and reshapes) collapses
this to 1 compute + 2 restores + 12 memo hits per row.

Quiet-machine A/B (pinned, optimized, `bench/lincmt_value_memo_ab.R`):

| cell | before (us/obs) | after | gain |
|---|---|---|---|
| 2-cmt 400x1000 | 8.51 | 5.59 | 1.52x |
| 2-cmt 40x200 | 7.50 | 4.00 | 1.88x |
| 3-cmt 400x1000 | 11.29 | 6.82 | 1.66x |
| 3-cmt 40x200 | 10.25 | 4.38 | 2.34x |

The 40x100 FOCEi fit cell moves 32.0 -> 30.2 s (warm reps 28.3 -> 26.4,
~1.07x; identical objective -11387.709011) -- the fit level is dominated
by costs outside the solve.  Forward sens columns match reverse mode to
7.2e-12; all 19 `test-lincmt*.R` files pass (6,617 assertions incl. the
new `test-lincmt-value-memo.R`); carry benches unchanged; the nlmixr2est
carry suite shows only the known cross-tree adr-default pair.

## Update: delta-keyed exponential memo (semigroup/backfill exploration outcome)

The user's "solve furthest and back fill" idea reduces, in its stable form, to the observation that the tail's exponentials depend only on (theta, gap): the theta window now carries a four-slot, delta-keyed memo of `exp(-L_i*delta)` / `exp(-ka*delta)` and their per-direction tangents, built with the exact fvar operation order so a hit is **bitwise identical** to recomputation (tested with `expect_identical` on/off). A give-up guard (eight consecutive misses) keeps no-reuse designs at parity, and re-arms on a row whose gap repeats the previous row's, so an irregular stretch no longer starves the regular rows that follow it under the same parameters (`51d740c6d`: a subject with fifteen irregular then a hundred regular observations went from 0 memo hits to 98; a fully irregular design still builds exactly 8 times, unchanged). `RX_LINCMT_DELTA_MEMO=off` and `rxode2:::linCmtDeltaMemo()` disable/force it; `linCmtSeqStats()` gains `expBuild`/`expHit`.

Measured (pinned, optimized, load-recorded, et-protocol; `bench/lincmt_delta_memo_ab.R`):

| cell | off µs/obs | on µs/obs | gain |
|---|---|---|---|
| 2-cmt uniform 400x1000 | 2.98 | 2.21 | **1.35x** |
| 3-cmt uniform 400x1000 | 5.00 | 4.65 | 1.07x |
| 2-cmt non-uniform 400x1000 | 1.84 | 1.88 | 0.98x (give-up) |
| 3-cmt non-uniform 400x1000 | 5.03 | 4.97 | 1.01x (give-up) |

Fit level (FOCEi 40x100 uniform, `bench/lincmt_delta_memo_fit_ab.R`): 30.0 -> 28.5 s (**1.05x**), objective identical to the last digit. Validation: memo tests 10/10 (bitwise, hit-rate mechanism, give-up, thread identity), all `test-lincmt*.R` 6,626/0, carry benches unchanged, nlmixr2est carry suite green (the known 2 cross-tree adr-default failures only). The literal backward backfill was shown numerically unusable and the closed-form adjoint structurally inapplicable to FOCEi (per-observation gradients consumed); the exploration record ships as `bench/lincmt_backfill_explore.R`.

## Direct state-read emission (P5)

The parser now registers derivative slots when `rx__sens_<cmt>_BY_<slot>` states are referenced as bare symbols, and `.rxD$linCmtB` emits concentration-gradient arithmetic over those states instead of `linCmtB(-2, .)` read calls (bitwise-identical, per-`trans` scaling mirrored exactly; uncovered `trans` keep the call form). A FOCEi inner model drops from ~11 `linCmtB` executions per row to the single value call; all rxode2 and nlmixr2est carry suites pass unchanged.

Measured honestly (commits d21ddc900 + the bench commit): the win is modest because read calls already short-circuited cheaply -- solve-level 1.0-1.12x, FOCEi 40x100 fit 40.39 -> 38.89 s (1.04x, identical objective). The remaining sensitivity-path cost is the value call's compute+restore executions and per-row solve machinery, not the reads.

## dydt/calc_lhs consolidation (thin value path)

In a fit each pure `linCmtB()` row is visited twice per inner evaluation: the solver's state fill (`linSolve` -> `dydt`, the computing entry) and the left-hand-side walk (`calc_lhs`, a restore re-execution). The value call now serves the second visit through a thin path -- saved amounts plus concentration scaling only, skipping the sensitivity setup, rate cache, Jacobian restore and concentration-gradient recompute -- with the Jacobian restored lazily if a carry sentinel or read call for that row follows (carry models unaffected, tested against reverse mode). Only the `linCmtB` entry changes; no solver driver is touched.

Mechanism (primary evidence, `linCmtSeqStats()`): a FOCEi posthoc's lhs walk is served entirely by the thin path (theo_sd probe: 274,204 `valueLite` executions vs 195,860 computes); the full linCmt suite (6,637 assertions), the carry benches and the nlmixr2est carry suites are green with thread bit-identity, and the posthoc objective is identical to the pre-change build. Wall-clock A/B was not resolvable above machine contention during measurement (loads 2.4-6.4 recorded in `bench/lincmt_value_lite_ab.R`'s results); the expected effect is small and positive, consistent with the P5-series findings that per-entry costs are minor once cached.

## Transition-matrix propagation of a repeated interval (`linCmtSensPhi`)

Where a sensitivity row's interval repeats -- regular sampling, and the dosing intervals of a repeated regimen -- that interval's state-transition matrix is assembled once and every later row of the same width propagates through it with plain multiply-adds, instead of evaluating the closed form again in each requested direction. The ingredients were already cached: the theta window holds the eigenvalues, spectral matrices and their tangents, and the delta-keyed memo holds the interval's exponentials.

**Engage rule.** A matrix pays for itself only when reused, and a delta-memo *hit* is exactly the evidence that this interval has occurred before under these parameters. So one is built on the first hit for an interval and reused after, and never on a miss -- a design whose intervals never repeat builds none at all and cannot be slowed by this path (measured: `phiBuild = 0`, all rows on the previous route). Rate-bearing infusion rows (affine rather than linear in the prior state) and steady-state rows keep the previous route, as before.

**Thread invariance.** Each subject starts from a blank interval state. The window is per *thread* and outlives a subject, so without that reset "has this interval been seen before" -- and with it which rows propagate through a matrix -- would depend on how subjects happened to be handed to threads, and the last digits of a solve would move with the thread count. Restarting per subject costs a few exponentials and one matrix per subject; a solve is identical however many threads run it (`expect_identical`, tested).

**These are both exact; the arithmetic order differs.** The matrix is summed first and then applied, where the previous route accumulated the same products as it went. Floating-point addition is not associative, so the two can differ in the last few digits -- neither is an approximation of the other and neither is the more correct. Over 70 cells (every compartment count, parameterization, regimen and direction mask) the largest disagreement is a few units in the last place of the values involved: elementwise it reads as 4.0e-10 at worst, but that figure is taken where a gradient column crosses zero -- in absolute terms the spread is 1.4e-12 against values of 2.0e+03, and 22 of the 70 cells are bitwise equal. Scored against an independently integrated reference (the equivalent ODE system at 1e-12, valid for the parameterization it encodes), the new order is the *closer* of the two on 63 rows against 20 with 13 tied -- so it is not the worse-conditioned order. `rxSolve(linCmtSensPhi=FALSE)` restores the previous order for digit-for-digit reproduction of earlier results.

**Measured** (`bench/lincmt_phi_engage_interleave.R`, one binary with the control flipped, arms alternated within each round so drifting load cancels; 20 subjects x 1000 observations, pinned):

| cell | engages | quieter run (load ~2.7) | contended run (load ~12) |
|---|---|---|---|
| 2-cmt uniform | yes | 1.22x | 1.12x |
| 3-cmt uniform | yes | 1.44x | 1.21x |
| 2-cmt multi-dose | yes | 1.05x | 1.06x |
| 3-cmt multi-dose | yes | 1.59x | 1.39x |
| 3-cmt infusion | no | 1.10x | 1.02x |

The infusion cell runs identical code in both arms (`phiRows = 0`), so it measures each run's noise floor -- 1.10x and 1.02x respectively. The gains at three compartments sit well clear of it in both runs; the 2-cmt multi-dose cell does not, and is reported as unresolved rather than as a gain. A separate build-to-build comparison against the pre-change tree found the non-engaging paths within that same noise.

**At the fit level** (`bench/lincmt_phi_fit_ab.R`, FOCEi, 40 subjects x 100 observations, one binary with the control flipped, arms alternated over 3 rounds, pinned, -O3): read the per-iteration column, not wall clock. The arms converge in different numbers of outer iterations -- up to 32% apart -- because a last-bit change in the gradient moves bobyqa's trajectory, and that swamps the timing.

| cell | wall (phiOff/phiOn) | per iteration | nIter on/off | reuse |
|---|---|---|---|---|
| 2-cmt uniform | 1.69x | **1.215x** | 936 / 1302 | 499:1 |
| 3-cmt uniform | 1.20x | **1.531x** | 1800 / 1410 | 499:1 |
| 2-cmt non-uniform | 0.80x | **1.052x** | 1434 / 1086 | 4:1 |

The per-iteration gains track the solve-level table above (1.22x, 1.44x); the wall-clock figures do not, and swing both ways on a single dataset -- so they are not evidence in either direction. The non-uniform cell is accordingly **not** a regression in work: it is marginally faster per iteration, and its 0.80x wall clock is entirely the 32% extra iterations. Objectives across arms agree to 4.6e-3 absolute on -1.9e+04 (2.4e-7 relative), inside the optimizer's own stopping spread -- the integrated arm sits the same distance away.

Against that integrated arm on the same build and dataset, the closed form is at wall-clock parity (21.7 s vs 22.7 s) while still costing 1.29x more per iteration (23.1 vs 18.0 ms) -- the parity leans on 936 iterations against 1263. These are not comparable to the older 28.5 s vs 14.3 s figures, which came from a different build, dataset and observation count.

### The engage rule had to count rows, not calls (fixed)

The fit-cell run surfaced a real fault: the non-uniform design built **zero** matrices in a plain `rxSolve()` yet **8.6M** inside a fit, at 4:1 "reuse". The cause is that one row reaches the row-Jacobian several times in a fit -- the generated model runs the value line from `dydt` and from `calc_lhs`, and the inner problem re-walks a subject many times -- and each execution looked its own gap up again. A row asking twice is the same interval, not a recurrence, but it was counted as one, so the give-up guard never disarmed and matrices were built for gaps that never repeat.

The evidence is now gathered per **row**: `lastIdx` gates the guard bookkeeping to genuinely new rows, and `deltaIdx` records which row cached a gap, so a hit from a *different* row can be told from that same row asking again. Only a cross-row hit re-arms the guard or engages a matrix; a re-query still reuses the cached exponentials, which is free and bitwise identical -- it simply is not evidence.

Re-measured at the **default** optimization (2 rounds, pinned, arms alternated):

| cell | arm | sec | nIter | ms/iter | phiBuild | reuse |
|---|---|---|---|---|---|---|
| 2-cmt uniform | phiOn | 28.40 | 1167 | 24.34 | 75,895 | 495:1 |
| 2-cmt uniform | phiOff | 45.36 | 1302 | 34.84 | 0 | - |
| 2-cmt uniform | ode | 23.05 | 1263 | 18.25 | 0 | - |
| 2-cmt non-uniform | phiOn | 37.86 | 1086 | 34.86 | **0** | - |
| 2-cmt non-uniform | phiOff | 37.98 | 1086 | 34.97 | 0 | - |

The non-uniform cell is now exactly inert: nothing is built in either arm, so the two run identical code and agree to the last digit *and* take the same 1086 iterations. Its previously reported 0.80x wall clock and 32% iteration divergence were artifacts of the spurious building, not properties of the design. Uniform still engages fully, and its per-iteration gain **survives at 1.43x** -- larger than the 1.22x reported above, which was measured at a forced `-O3`; at the default optimization an installed package actually uses, the fvar tail the matrix replaces is relatively more expensive. Against the integrated arm the closed form remains 1.33x more expensive per unit work (24.34 vs 18.25 ms/iter).

A plain solve queries each row exactly once (`expHit == 0`), so this fault was not reachable from rxode2's own test suite -- reproducing it needs a fit, and rxode2 does not depend on nlmixr2est. `bench/lincmt_phi_fit_engage_check.R` is the committed fit-level reproducer (asserts zero builds on a non-repeating design and sustained reuse on a repeating one); the definitive regression test belongs in nlmixr2est's suite.

**Validation.** New `test-lincmt-phi-engage.R` (20 assertions: engage rule both ways, round-off agreement across compartment counts and regimens, thread identity under the default). All 22 `test-lincmt*.R` files, 6,668 assertions, 0 failures. Carry benches unchanged (3.078e-11; 11/11). nlmixr2est carry suites 186 pass with only the 2 known cross-tree `adr-default` failures. `test-lincmt-delta-memo.R` now pins `linCmtSensPhi="off"` so it continues to test the memo's own bitwise exactness against itself rather than against a different operation order; its bitwise assertions are kept, not loosened.

Not done: `which1=-4` (the carry's state-transition entry) is mathematically an entry of this matrix, but the carry exists for time-varying covariates -- exactly the case where parameters change per row, the window refills and no interval is ever reused -- so the cache is essentially never populated when that sentinel is called, and it deliberately recomputes from its own arguments to stay independent of cached state. Left alone.


mattfidler:
Superseded by #1288. GitHub closed this PR automatically when its base branch `lincmt-hybrid` was deleted on merging #1281, and refuses to reopen it against a deleted base. #1288 has the same head branch and commits, now based on `main`.

PR #1285, opened 2026-08-24
Size the individual solve pool by nsim*nsub (fixes nlmixr2#412 'nothing to solve')

mattfidler:
## The report

nlmixr2/nlmixr2#412: `addNpde()` and `vpcSim()` on a two compartment
`linCmt()` FOCEi fit both stop with

```
Error in rxSolveSEXP(object, .ctl, .nms, .xtra, params, .eventsForSolve,  :
  nothing to solve
```

while everything else about the fit works.

## Root cause

`rxSolve_datSetupHmax()` (`src/rxData.cpp`) sizes the per-individual solve
pool (`inds_global`, via `rxOptionsIniEnsure()`):

```c
if (nid == nPopPar || nPopPar == 1) {
  ntot = nid;
} else {
  ntot = nid*nPopPar;   // <-- nsub times too many
}
```

A solve only ever needs one `rx_solving_options_ind` per individual solve,
which is `rx->nsub * rx->nsim`. That is exactly `nPopPar`: a few lines later
rxode2 itself derives `rx->nsim = nPopPar / rx->nsub` and requires `nPopPar`
to be a multiple of `nsub`, and `rxSerialize.cpp` allocates the same pool as
`nsub * rx->nsim`.

Multiplying by `nid` on top of that makes the pool grow with the *square* of
the number of subjects. With `nsim = 300` (the `vpcSim()`/`addNpde()`
default) on the reported linCmt model:

| subjects | pool asked for | outcome (before) |
|---|---|---|
| 12 (theo_sd) | 43,200 | works |
| 400 | 4.8e7 (47 GB) | works only because `calloc` is lazy |
| 1000 | 3.0e8 (297 GB) | `'R_Calloc' could not allocate memory (300000000 of 992 bytes)` |
| >= ~2700 | overflows `int` -> negative | **`nothing to solve`** (`if (ntot <= 0) stop(...)`) |

The `ntot <= 0` guard right below is what turns the overflow into the
reported message.

## Fix

Size the pool as the number of individual solves:

```c
if (nPopPar == 0) {
  ntot = 0;              // keep the existing 0-row-params guard
} else if (nid > nPopPar) {
  ntot = nid;
} else {
  ntot = nPopPar;
}
```

Results are unchanged -- only the allocation size differs. Verified byte
identical `sim` output before/after for the four parameter shapes that reach
this branch: `nsim` with multi-subject data, `nPopPar == 1`, `nSub` with
single-subject data (`simSubjects`), and a `params` data.frame of
`nsub * nsim` rows.

## Effect

The reproducer from the issue (2700 subjects, 2 compartment `linCmt()`,
`est = "focei"`):

```
before:  addNpde ERROR: nothing to solve
         vpcSim  ERROR: nothing to solve
after:   addNpde OK; NPDE range: -2.935 .. 2.935
         vpcSim  OK rows: 2430000
```

`rxSolve(..., nsim = 300)` on 2700 subjects now takes 2.3 s / 1.9 GB peak.

## Tests

`tests/testthat/test-nsim-ind-alloc-412.R`

- 33,000 subjects x `nsim = 2` puts the old size (`33000 * 66000`) past
  `INT_MAX` while the real pool is 66,000. Confirmed to fail with exactly
  `nothing to solve` on the unpatched build and to pass here (~1 s, ~140 MB).
- a `params` data.frame of `nsub * nsim` rows solves as `nsim` replicates,
  each matching the corresponding single-parameter-set solve -- this is the
  `nid != nPopPar` branch that changed.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -278,6 +278,14 @@ mod |> ini(prior(eta.cl, eta.v) ~ invWishart(4))
 
 ## Bug fixes
 
+- A multi-subject `rxSolve()` with `nsim`/`nStud > 1` no longer sizes the
+  per-individual solve pool as `nsub` times the number of individual solves
+  it needs.  The over-allocation grew with the square of the number of
+  subjects, so a large study either ran out of memory or overflowed the size
+  to a negative number and stopped with `nothing to solve` -- which is what
+  made `nlmixr2est::addNpde()` and `vpcSim()` fail on a large fit
+  (nlmixr2/nlmixr2#412).  Results are unchanged.
+
 - A chunked solve (`rxSolve(file=`/`chunkSize=`)) with `nStud > 1` now
   simulates the omega uncertainty it was asked for.  It previously returned a
   plausible looking result drawn entirely from the point estimate omega, with
--- a/src/rxData.cpp
+++ b/src/rxData.cpp
@@ -4253,7 +4253,9 @@ static inline void rxSolve_datSetupHmax(const RObject &obj, const List &rxContro
       // one solve per subject).  Multiplying nid by nPopPar over-allocated by a
       // factor of nsub: a large study then either exhausted memory or overflowed
       // `int` and reported "nothing to solve" (nlmixr2/nlmixr2#412).
-      if (rxSolveDat->nPopPar == 0) {
+      if (nid == 0 || rxSolveDat->nPopPar == 0) {
+        // both were 0 under the old expression too, so the ntot <= 0 guard
+        // below keeps firing exactly where it used to
         ntot = 0;
       } else if (nid > rxSolveDat->nPopPar) {
         ntot = nid;
--- a/src/rxData.cpp
+++ b/src/rxData.cpp
@@ -4247,12 +4247,18 @@ static inline void rxSolve_datSetupHmax(const RObject &obj, const List &rxContro
           }
         }
       }
-      // if (nid == 0){
-      // } else
-      if (nid == rxSolveDat->nPopPar || rxSolveDat->nPopPar == 1){
+      // One `rx_solving_options_ind` is needed per individual SOLVE, that is
+      // `nsub*nsim`, which is `nPopPar` whenever the parameters were expanded
+      // per subject (`nPopPar == 1` is a single population set, i.e. nsim=1 and
+      // one solve per subject).  Multiplying nid by nPopPar over-allocated by a
+      // factor of nsub: a large study then either exhausted memory or overflowed
+      // `int` and reported "nothing to solve" (nlmixr2/nlmixr2#412).
+      if (rxSolveDat->nPopPar == 0) {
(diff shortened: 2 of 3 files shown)


mattfidler:
### Verification

**Full rxode2 test suite** (`NOT_CRAN=true`, 289 files): **38,272 passing, 0
failures, 0 errors**, 2 skipped.

**End to end on the reported case** (2700 subjects, two compartment `linCmt()`,
`est = "focei"`):

```
before:  addNpde ERROR: nothing to solve      after:  addNpde OK; NPDE range: -2.935 .. 2.935
         vpcSim  ERROR: nothing to solve              vpcSim  OK rows: 2430000
```

`rxSolve(..., nsim = 300)` on 2700 subjects: 3.6 s, 1.9 GB peak.

**Results unchanged:** byte identical `sim` output before/after for the four
parameter shapes that reach this branch -- `nsim` with multi-subject data,
`nPopPar == 1`, `nSub` with single-subject data (`simSubjects`), and a `params`
data.frame of `nsub * nsim` rows.

### Review

Reviewed independently by Antigravity (gemini-3.1-pro-high). Three findings,
triaged by measurement:

1. **Acted on.** Gating only on `nPopPar == 0` dropped the `ntot <= 0` guard
   for a 0-subject (`nid == 0`) event table -- the old `nid * nPopPar`
   expression came out 0 whenever *either* factor was 0. Second commit gates on
   both and pins it with a test. Today an empty data set is caught earlier by
   etTrans's "no rows in event table or input data", so this is defensive.
2. **Rejected.** "A first `ID` of NA undercounts `nid`." Unreachable: etTrans
   rejects the input first with `'id' cannot have NA values`, on both the
   patched and unpatched builds.
3. **Rejected as out of scope, but note it is real.** "A 1-row `params`
   data.frame with a multi-subject data set leaves later subjects
   unpopulated." Measured identical on the patched and unpatched builds --
   `rxSolve(m, data.frame(cl = 1.5), <2 subjects>)` returns `NA` for subject 2
   either way. `rxSolve_assignGpars()` loops to `nPopPar` (1) and never
   broadcasts that row to the remaining subjects, while the
   `nPopPar % nsub != 0` guard exempts `nPopPar == 1`. Pre-existing and
   untouched by this diff; worth its own issue.

mattfidler:
Two further Antigravity rounds on the updated diff:

- **Round 3** raised only that the `nStud > 1` path was covered implicitly
  (`nsim` resolves to `nStud` for a data.frame event table) rather than
  asserted. Added an explicit `nStud = 3` test.
- **Round 4** came back clean: no correctness, memory-safety or concurrency
  defect introduced, having walked `rxHomGroups`, `simSubjects`, the chunked
  (`chunkSize=`) path, a `params` data.frame, `nPopPar == 0` and the
  `nsim`/`nStud` combinations against the `nsub * nsim` pool invariant.

PR #1286, opened 2026-08-25
Allow a marginal prior on an omega covariance (off-diagonal) element

mattfidler:
## Summary

`rxPriorLogDensity()`/`rxPriorBuildSpec()` can now evaluate a `dnorm()`/`dcauchy()` prior placed directly on ONE omega covariance cell, not just a diagonal (variance) one -- `prior(eta.cl, eta.v) ~ dnorm(0, 0.1)` on a model whose `eta.cl`/`eta.v` covary. Distinct from a whole-block distribution (`invWishart()`/`multiNormal()`), which still requires naming the model's entire connected block -- a marginal prior only needs its two names to covary, so it also works on a two-name subset of a larger block. NONMEM does not reject this and neither should this kernel; it was previously refused outright by a hard `stop()`, unreachable through real syntax until lotri's own `prior()` parsing was relaxed to allow it (nlmixr2/lotri#59).

## Implementation

- `rx_prior_term_t` gets a new `etaIdx2` field (`inst/include/rxode2prior.h`), additive and backward compatible -- `0` for every existing term (population parameter or diagonal omega element, unchanged), nonzero addresses the off-diagonal pair `(etaIdx-1, etaIdx2-1)`.
- `termValue()`/`addGrad()` in `src/priorDensity.cpp` are the only C++ evaluator changes; `evalInvWishartTerm()` (whole-block, type 3/4) never uses them and is untouched.
- **Caught during testing via central difference**: `addGrad()`'s off-diagonal branch must write ONLY the one cell `termValue()` actually read, matching the kernel's own documented entrywise `gradOmega` convention -- writing the same scalar into both symmetric cells (which seemed to mirror invWishart's own naturally-symmetric gradient) measured exactly 2x the correct total, since a downstream caller that wants the derivative along the "move the pair together" direction sums both cells itself.
- `R/prior-sim.R`'s `usePrior=TRUE` simulation path does not yet support this (it draws a deviation added to the model's own initial estimate, a different mechanism) and now gives an explicit, accurate error instead of a confusing "unknown omega element" fallthrough its existing eligibility filter would otherwise have produced once the off-diagonal row became reachable.

## Test plan

- [x] New tests in `tests/testthat/test-prior-density.R`: real `ini()`/`prior()` syntax reaches the off-diagonal row, `rxUiPriors()` surfaces it, central-difference gradient check, whole-block `invWishart()` unaffected, `prior-sim` guard gives a clear error.
- [x] Updated the one existing test that asserted the old (now-relaxed) rejection behavior.
- [x] Full prior-related test suite (`test-prior-density.R`, `test-assert-priors.R`, `test-ini-prior-column.R`, `test-ini-prior-piping.R`, `test-prior-sim-*.R`) passes with no regressions.

Depends on nlmixr2/lotri#59. Part of the shared kernel work for nlmixr2/nlmixr2est#929 (companion PR lands in nlmixr2est, requiring zero source changes there).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -141,8 +141,6 @@ Description: Facilities for running simulations from ordinary
     'CVODE' sources and headers are released under the BSD-3-Clause license.
     The information is available in the inst/COPYRIGHTS.
 BugReports: https://github.com/nlmixr2/rxode2/issues/
-Remotes:
-    nlmixr2/lotri@offdiag-cov-prior
 NeedsCompilation: yes
 VignetteBuilder: knitr
 License: GPL (>= 3)
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -141,6 +141,8 @@ Description: Facilities for running simulations from ordinary
     'CVODE' sources and headers are released under the BSD-3-Clause license.
     The information is available in the inst/COPYRIGHTS.
 BugReports: https://github.com/nlmixr2/rxode2/issues/
+Remotes:
+    nlmixr2/lotri@offdiag-cov-prior
 NeedsCompilation: yes
 VignetteBuilder: knitr
 License: GPL (>= 3)
--- a/NEWS.md
+++ b/NEWS.md
@@ -47,6 +47,23 @@
   `rxPriorOmegaToCholOmegaInvGrad()` C++ function is pure, with no R/Rcpp
   call, and is exposed through the same function-pointer table.
 
+- `rxPriorLogDensity()`/`rxPriorBuildSpec()` now evaluate a marginal
+  `dnorm()`/`dcauchy()` prior placed directly on ONE omega covariance
+  (off-diagonal) element, not just a diagonal (variance) one --
+  `prior(eta.cl, eta.v) ~ dnorm(0, 0.1)` in a model whose `eta.cl`/`eta.v`
+  covary. This is distinct from a whole-block distribution
+  (`invWishart()`/`multiNormal()`), which still requires naming the
+  model's entire connected block; a marginal prior only needs its two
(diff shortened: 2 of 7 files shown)


mattfidler:
An independent review (\`agy\`/Gemini) found a real memory-safety gap: \`.rxPriorKeyIndex()\`'s new off-diagonal branch read \`neta1\`/\`neta2\` straight off the matched \`iniDf\` row with no validation, and \`src/priorDensity.cpp\`'s \`termValue()\`/\`addGrad()\` index \`omega[e1][e2]\` with no bounds check of their own (they run inside an OpenMP-parallel region and cannot safely call back into R to error). A hand-edited or unusually-piped \`iniDf\` whose off-diagonal row references an eta index with no corresponding diagonal row would read/write out of bounds.

Fixed in ad954d963: the R-side spec-build step now validates both indices correspond to real diagonal etas before the C kernel ever sees them, with a regression test. Full prior test suite (120+ tests across \`test-prior-density.R\` and 8 other prior-related files) still passes.

mattfidler:
All CI jobs were failing: this PR's tests exercise \`prior(a, b) ~ dnorm(...)\`, but CI installs lotri from its normal source (CRAN/main), which doesn't have that support yet -- every job failed on \`prior 'dnorm' is univariate and cannot be applied to the block ...\`, exactly the lotri-side rejection nlmixr2/lotri#59 removes.

Added a temporary \`Remotes: nlmixr2/lotri@offdiag-cov-prior\` pin in 7ec0c56 so CI installs the fixed lotri branch. Will remove it once nlmixr2/lotri#59 merges (a stale \`Remotes:\` pin would eventually surface as a CRAN check NOTE).

Issue #1289, opened 2026-08-25
`rxSolve()` on an nlmixr2 fit object leaks ~2.4 MB per call and re-lowers the model on every call


billdenney:
## Summary

Calling `rxode2::rxSolve()` repeatedly on an `nlmixr2` fit object grows process
memory by ~2.4 MB per call without bound. The growth is invisible to `gc()` —
the R heap stays flat — and is not released by `rxUnloadAll()` or `rxClean()`.

The cause appears to be that `rxSolve.nlmixr2FitData` re-derives the lowered
`rxode2` model from the fit on **every** call (via `$simulationModel` →
`rxode2()` → `rxGetModel`) rather than caching it on the fit. Lowering the fit
once and solving the resulting plain `rxode2` model is ~16x faster and does not
grow.

This only reproduces for ODE models. An analytic (`linCmt()`) fit shows neither
the per-call cost nor the growth.

## Reprex

Self-contained; uses `nlmixr2data::theo_sd`. The `ps` package is used only to
read process memory.

```r
suppressPackageStartupMessages({
  library(nlmixr2)
  library(rxode2)
})

# Process private bytes minus whatever the R heap is holding.  gc() is forced
# first so R-level allocation cannot account for the difference.
mem_mb  <- function() {
  invisible(gc(FALSE))
  unname(ps::ps_memory_info()[["mem_private"]]) / 1048576
}
heap_mb <- function() sum(gc(FALSE)[, 2])

one.cmt <- function() {
  ini({
    tka <- 0.45
    tcl <- 1
    tv  <- 3.45
    etaKa ~ 0.6
    etaCl ~ 0.3
    etaV  ~ 0.1
    addSd <- 0.7
  })
  model({
    ka <- exp(tka + etaKa)
    cl <- exp(tcl + etaCl)
    v  <- exp(tv  + etaV)
    d/dt(depot)  <- -ka * depot
    d/dt(center) <-  ka * depot - cl / v * center
    cp <- center / v
    cp ~ add(addSd)
  })
}

fit <- nlmixr2(one.cmt, nlmixr2data::theo_sd, est = "saem",
               control = saemControl(print = 0, nBurn = 20, nEm = 30))

ev <- rxode2::et(amt = 320, cmt = "depot") |> rxode2::et(seq(0, 24, by = 0.5))

bench <- function(label, obj, n = 60) {
  invisible(rxode2::rxSolve(obj, ev))          # warm up outside the measurement
[...]

PR #1290, opened 2026-08-25
fix(parse): stop leaking memory on every model translation

billdenney:
Translating a model that has already been translated grew process memory
without bound, invisibly to `gc()`, `rxUnloadAll()` and `rxClean()`. This is
the per-lowering cost underneath #1289 — `rxSolve()` on a nlmixr2 fit
re-derived the model on every call, and each of those re-derivations leaked.

Reprex on `main`:

```r
library(rxode2)
rss <- function() { invisible(gc(FALSE)); unname(ps::ps_memory_info()[["rss"]]) / 1048576 }
m <- "d/dt(a) = -k*a;"
invisible(rxNorm(m))
before <- rss(); for (i in 1:500) invisible(rxNorm(m)); rss() - before
#> 192.4    ... MB, with the R heap flat; rxUnloadAll()/rxClean() do not give it back
```

Three independent causes, found by counting memoise keys and `gc()` cells, and
by valgrind for the native one.

### 1. `.rxModelVarsCharacter()` translated through the memoised path

The prefix it passes was derived from `tempfile()`, so it differed on every
call. `rxTrans.character()` is memoised and memoise keys on every argument, so
each call **missed the cache and added an entry to it** — two entries and
~0.4 MB per `rxModelVars()`/`rxNorm()` call, kept for the life of the session.

The fix is not to make that cache hit. Translating is not only a computation:
it leaves the parsed model in the C parser's state, and this function's callers
depend on that state — code generation, model loading, model piping and
`modelDesc` all read what the parser last saw. (I tried the cache-hit version
first; the suite caught it — `test-load.R` "nothing in output queue to write"
after `rxDelete()` + `$compile()`, plus `test-lincmt-modelvars.R`,
`test-lincmt-neg-depot-1275.R`, `test-ui-dv.R` and `test-ui.R:806`.)

So the memoised `rxTrans.character()` is now a thin wrapper over a plain
`.rxTransCharacter()`, and `.rxModelVarsCharacter()` calls that directly: it
always translates for real, exactly as before, and never feeds the cache. The
prefix is also derived from the model rather than from `tempfile()`, so
repeated translations of one model stop minting a new name each time (R interns
names for the life of the session).

### 2. `tb.lho` was allocated and never freed

`reset()` (src/tran.c) `R_Calloc`s `tb.lho` immediately after `tb.lh`, but
`parseFree()` freed only `tb.lh`. Every parse leaked `MXSYM * sizeof(int)` =
200 KB. valgrind names it exactly:

```
200,000 bytes in 1 blocks are definitely lost in loss record 5,941 of 6,045
   by 0x...: reset (tran.c:392)
   by 0x...: setupTrans (tran.c:697)
   by 0x...: _rxode2_trans (tran.c:771)
```

Diffing what `reset()` allocates against what `parseFree()` frees showed
`tb.lho` as the only member of the first list missing from the second.

### 3. The user-function search list was unbounded

`.udfAddToSearch()` appended the calling environment to `.udfEnv$searchList`
and never pruned it, so every `$` on a `rxUi`, every `rxSolve()` and every
`rxode2()` pinned a call frame — usually one that was already dead — for the
rest of the session. Its index made it worse: an environment keyed by
`data.table::address()`, minting a new name per environment.

The list is now bounded (`options(rxode2.udfSearchLimit=)`, default 20, oldest
forgotten first — a long-lived environment such as the global environment is
simply re-added the next time it is used) and membership is tested with
`identical()`, which needs no names at all. Search order is unchanged.

### Result

Steady-state growth per call, same model each time (50 calls per block, RSS and
`gc()` cells measured after a forced full `gc()`):

| | before | after |
| --- | --- | --- |
| `rxNorm(m)` | +0.385 MB rss | 0.000 |
| `rxode2(m)` | +0.191 MB rss | 0.000 |
| `getBaseSimModel(ui)` | +0.025 MB cells | 0.000 |
| `ui$theta` | +0.005 MB cells | 0.000 |

### Tests

`tests/testthat/test-memory-growth.R`. The invariants that make the leaks
impossible run by default:

* repeated translation of one model does not grow the memoise cache;
* the parse prefix is a function of the model (same model → same prefix,
  different model → different prefix);
* a model given as text always leaves the parser holding *that* model, and a
  model whose C file was deleted can still regenerate it;
* the search list stays bounded, honours the option, and falls back on a bad
  option value.

The three tests that measure process memory are opt-in — they are slow and RSS
can move underneath them on a loaded machine:

```r
Sys.setenv(RXODE2_MEMORY_TEST = "true")
```

They measure steady-state growth after a warm-up loop, against thresholds far
below what the leaks produced: <20 MB over 500 `rxNorm()` calls (was ~190 MB),
<20 MB over 300 `rxode2()` calls (was ~57 MB), <5 MB of cons cells over 500
`getBaseSimModel()` calls (was ~12 MB). `ps` is added to `Suggests` for them.

### Verification

Full suite, this branch vs `origin/main` built the same way: baseline
`[ FAIL 27 | PASS 38989 ]`, and this branch reaches the same failure set (the
27 pre-existing failures are unrelated — `lotri` prior distributions, symengine
`rxFromSE()` derivatives, a queue error in `test-load.R`).


mattfidler:
Thanks @billdenney

PR #1304, opened 2026-08-27
linCmt(): analytic gradient for the sensitivity row Jacobian

mattfidler:
A `linCmt()` sensitivity row is affine in the prior state — read straight off
`linCmtStanNTail()`, `ret = Phi(dt) yp + b(dt)` — and both `Phi` and `b`, with
all their parameter tangents, have a closed form in the constants the
theta-keyed window already holds: the eigenvalues, the spectral matrices, `ka`,
and each of their derivatives. Two changes follow, measured together.

## Closed-form transition matrix — `linCmtSensPhi = 2`, now what `TRUE` means

Until now `Phi` was obtained by **probing** the kernel with unit-basis prior
states: exact by construction, but one kernel evaluation per direction per
column. At that price it could only be paid on an interval shown to recur, and
rate-bearing rows had to be excluded. Assembling it from its closed form costs
about one kernel evaluation, so it is available for an interval that never
recurs, and `b` carries the depot and infusion terms.

`Phi` does not depend on the rates, so it is still cached per interval exactly
as the probed one was — where intervals repeat, that reuse is worth more than
any build. An earlier version of this that rebuilt per row **lost** to the
probe on regular designs, which is what forced the caching design. What is new
is the rows the probe declined.

## Multi-direction forward mode — `linCmtSensType = "ADm"`

`src/linCmtDualN.h` adds a dual scalar carrying one value and N tangents, so
the closed form is evaluated once for all directions instead of once per
direction. Unlike the window path it is not restricted to ordinary rows: the
steady-state kernels, which have no constants/tail factorization, get the same
sharing. It is an **option, not the default** — it wins at two compartments and
loses at three, where the wider scalar spills through Eigen's 3×3 products.

## Exactness

- **`"ADm"` is bitwise identical to `"AD"`** over 2520 cases: 1/2/3 compartments
  × IV/oral × every `trans` × all four steady-state forms × infusion × `dt` ×
  every requested-direction mask. That required giving `dualN` *fvar's* Eigen
  cost traits rather than costs scaled by the tangent count — Eigen uses them to
  choose how far to unroll a small product, and a different unrolling
  reassociates the sum.
- The closed-form matrix is a summation-order change, as the probed one already
  was. It agrees with the probed matrix (exact by construction) to **1e-14**,
  with the row tail to 2e-14, and its distance from reverse mode **is** the row
  tail's own distance. A FOCEi objective is unchanged to 1e-11.
- `linCmtSensPhi = FALSE` still reproduces the earlier row-by-row order digit
  for digit.

## Speed

Optimized build, single thread pinned. **1.19x–2.22x** over the previous default
at the solve level, and it carries into a fit: a 40 subject × 100 observation
2-cmt oral FOCEi fit runs **1.37x** faster on an irregular schedule (53.4 s →
38.9 s) and unchanged on a uniform one.

Which designs those are is the whole of it — where sampling is regular the
probe already served nearly every row and nothing changes. Worth noting: under
the probe, the *irregular* design was the more expensive of the two, its
bookkeeping paid on rows that never got a matrix. That inversion is gone.

**The manuscript's NONMEM comparison is measured on a uniform design** (a single
0.5 h gap across all 99 intervals), so re-running it under both routes moves the
per-direction slope by under 1% — the right answer, not a disappointing one. A
new `design-regularity-kit` in the write-up repo varies the schedule instead and
reports each program against its own uniform cell.

## Verification

- Full suite **38,690 pass / 0 fail / 0 error**, re-verified after merging
  `origin/main`.
- New: `tests/testthat/test-lincmt-dual.R`,
  `tests/testthat/test-lincmt-phi-analytic.R`.
- `tests/testthat/test-lincmt-phi-engage.R` covers the *probe's* engage rule, so
  it now names `linCmtSensPhi = 1` and gains a case for the closed form on the
  design the probe declines.
- Harnesses: `bench/lincmt_dual_valid.R`, `bench/lincmt_phi_analytic_valid.R`,
  `bench/lincmt_analytic_ab.R`, `bench/lincmt_focei_phi_ab.R`,
  `bench/lincmt_focei_phi_validate.R`.

## Notes for review

- No new `rxControl` slot, no pointer-table entry, no `$predDf`/`$iniDf` change.
  `"ADm"` is an added integer in the existing `linCmtSensType` slot;
  `linCmtSensPhi` gains levels within its existing one.
- `linCmtSensType` default is still `"auto"`, which still resolves to `"AD"`.
  The analytic gradient is reached through `linCmtSensPhi`, not through the
  sensitivity type, so a default solve gets it.
- The first commit is a separate, paths-only fix: seventeen bench scripts
  `load_all()` from three worktrees that no longer exist.


mattfidler:
## The NONMEM comparison has now been run

The PR body noted that the manuscript's NONMEM comparison sits on a uniform
design and so cannot see this change. The `design-regularity-kit` was built to
vary the schedule instead, and both arms have now run on the same committed
datasets — NONMEM 7.x ADVAN4 TRANS4 on the licensed machine, nlmixr2 here.

Irregular schedule against uniform, **each program against itself** (which is
what makes it comparable across two machines):

| | irregular / uniform |
|---|---|
| **NONMEM (ADVAN4 TRANS4)** | **0.96x** |
| `linCmtSensPhi = 0` — row by row | 1.11x |
| `linCmtSensPhi = 1` — probe-built matrix *(the previous default)* | **1.68x** |
| `linCmtSensPhi = 2` — closed-form matrix *(this PR's default)* | **1.23x** |

**NONMEM is flat.** Whatever ADVAN4 does per interval, it neither cheapens when
intervals repeat nor dears when they do not — so the cost of irregularity was
ours, not the method's.

The probe-built matrix paid 1.68x, and on that design it was *slower than
building no matrix at all* (0.365 vs 0.352 s/evaluation): its bookkeeping is
charged against rows that never get a matrix. Assembling the same matrix from
its closed form brings that to 1.23x — about two thirds of the excess over
NONMEM. **A residual remains, and the kit does not explain it.**

One control worth recording: the four-width `stretches` design comes out
*cheaper* than uniform in **both** programs (0.83x NONMEM, 0.95x here). Both
move together, so that is a property of rich early sampling over short
intervals rather than of either implementation — which is the agreement a
design-varying comparison needs before its disagreements are worth believing.

Within nlmixr2 the three routes agree to 3e-10 on an objective of order 1e4.

Kit, data and results: nlmixr2/linCmt-time-varying@8a9e62a.

mattfidler:
### Where the residual 1.23x was coming from -- and it is mostly gone

Follow-up to the closed-form transition matrix. The irregular/uniform ratio the design-regularity kit reported was **not arithmetic**: forcing the delta memo off (`RX_LINCMT_DELTA_MEMO=off`) makes an irregular schedule and a uniform one cost the same to within a percent under every `linCmtSensPhi` route. The per-row work does not depend on the schedule.

What the ratio measured was a discount only a regular design could collect -- plus a small charge the irregular one paid for trying. The give-up guard stops the window caching gaps that never recur, which is right; but one row is looked up several times under one theta (once per generated `linCmtB()` call, and again on every inner re-walk of a fit), and the guard was discarding those repeats too. After it disarmed, an irregular design reassembled the same transition matrix **once per execution** rather than once per row -- 8 builds and 32 hits per theta window instead of 100 and 400.

`730b08b83` gives the guard a slot outside the round-robin holding the current row's gap. It is never scanned by the speculative cache and never reports `crossRow`, so the guard's reading of the design, and the probe-built route's engage rule, are unchanged.

**Matched A/B** -- two rxode2 builds installed into separate libraries and alternated, one nlmixr2est, one budget, pinned; 40 subjects x 100 observations, 2-cmt oral, 3 fixed-omega etas:

| route | before | after |
|---|---|---|
| `linCmtSensPhi = 0` (row by row) | 1.15x | 0.95x |
| `linCmtSensPhi = 1` (probed matrix) | 1.74x | 1.46x |
| `linCmtSensPhi = 2` (closed form, default) | **1.30x** | **1.09x** |
| NONMEM ADVAN4 TRANS4 | 0.96x | 0.96x |

The irregular fit itself is 1.14x faster; uniform and the four-width design are unchanged, and a single-pass solve shows no regression either way. What remains at 1.09x is the assembly a distinct interval genuinely needs and a uniform design amortizes over the subject -- not cacheable, because there is nothing to cache.

Results are **bitwise identical** either way (the memo is exact caching), tested with it forced on and off across 1/2/3 compartments and both matrix routes. `linCmtSeqStats()` reports the builds served this way as `expSolo` (appended). New harness: `bench/lincmt_regularity_memo_ab.R`.

Full suite **39,649 pass, 0 fail**.

Written up in the paper repo (`nlmixr2/linCmt-time-varying@616e1c5`), which also refreshes the gradient-slope sweep from the re-run NONMEM arm: per-direction 0.00564 -> 0.00590, so the headline is 1.08x rather than 1.13x, base ratio 1.81x, and NONMEM's eta sweep is now reported as failing the kit's own curvature test (p 0.033) so its slope reads as an average over 2-5 directions.

PR #1309, opened 2026-08-28
perf: remove the dparser ambiguity that made parsing quadratic; translate symengine text in C

mattfidler:
## What this does

Two independent wins, plus the tests and tooling that pin them.

### 1. Parsing is no longer quadratic in model size

`inst/tran.g` was ambiguous. `statement`'s last alternative was a bare
`end_statement`, and `end_statement : (';')*` is nullable -- so `statement`
derived the empty string, and `statement_list : (statement)+` admitted any
number of empty statements at every position. dparser resolved that by
greediness, and `cmp_greediness`/`get_all` re-walk the whole accumulated parse
tree at each position: `perf` put ~90% of a parse in
`set_add`/`get_all`/`set_union`/`set_find`, with `get_all` recursing once per
statement.

Every parse in the package paid for it -- `rxode2()`, `rxNorm()`,
`rxModelVars()`, `rxS()`, `rxOptExpr()`. Requiring one `';'` in that
alternative makes `statement` non-nullable and the grammar unambiguous.

### 2. `rxNorm()` parses once instead of twice, and `rxOptExpr()` once instead of three times

Tracing `.rxTransCharacter` through one `rxOptExpr()` found three parses of the
same text, none of which was the counting pass or the replacing pass (those are
R-level walks over `quote({...})` and cost no parse at all):

1. `rxOptExpr` -> `rxModelVars(x)` for `.mv$params`, which was assigned and
   never read. Dead.
2. `rxNorm(x)` -> `rxCondition(x)` -> `rxNorm(x, FALSE)`.
3. `rxNorm(x)` -> `rxModelVars(x)`, the value it returns.

2 and 3 are the same computation twice: `rxCondition()`'s lookup key is a
digest **of the normalized model**, which is also what `rxNorm()` returns when
no condition is set. `.rxConditionKey()` is split out so `rxNorm()` can look the
condition up with the text it is about to return.

### Measured, 301-line model

| | before | after |
|---|---|---|
| `rxNorm()` | 3.179 s | **0.039 s** (82x) |
| `rxOptExpr(chunkLines = 0)` | 5.613 s | **0.335 s** (17x) |

Per-line parse cost grew 9.8x between a 25-line and a 301-line model; it is now
flat. `tests/testthat/test-parse-empty-statement.R` pins that shape so the
ambiguity cannot come back unnoticed.

### 3. symengine text translation in C

`rxFromSE()`/`rxToSE()` were 89.8% of symbolic-derivative setup while
`symengine::D()` was 0.46%. Two dparser grammars (`inst/seFromSE.g`,
`inst/rxToSE.g`) and C emitters translate directly; anything an emitter will
not reproduce byte-for-byte returns `NA_character_` and the R walker runs
instead. Jacobian and sensitivity builds hand the whole set over as one batch,
which is spread across threads above 250 expressions. `.rxSens()` on a
fifteen-state model: 1.27 s -> 0.19 s.

### 4. The common-subexpression search in C

`rxOptExpr()` counted subexpressions in a named R list and looked them up with
`[[text]]`, a linear scan, so the search was O(k^2) in the number of distinct
subexpressions. On a second-order sensitivity model k is enormous and that scan
is where a large model spent its time. `inst/rxCse.g` plus a C driver
(`src/rxCse.cpp`, `src/rxCseRun.h` and the `rxCse*.h` machines) hash the map and
run the per-statement walks under OpenMP. Output is byte-identical or the call
declines -- all or nothing, because a partial result cannot be mixed with R's
global counts.

| model | `rxNorm()` | CSE (R) | CSE (C) |
|---|---|---|---|
| wbc-shaped, 260 lines, 70 KB | 0.35 s | ~2.9 s | **0.04 s** |
| nimo-shaped, 286 lines, 3.1 MB | 59.5 s | ~143 s | **1.53 s** |

The search itself is ~93x faster; end-to-end on nimo it is 3.4x, because
parsing now dominates what is left.

Three things were subtle enough to be worth naming. There are **two** renderers,
one building the counting keys and one the output text, and they disagree
(`1*x` is counted as `1*x` and rendered as `x`), so they cannot be merged. `^`
evaluates its **right** operand first, because `.rxOptBin` tests
`as.numeric(e2)` before touching `e1`, and that decides the naming tie-break
between two candidates of equal `nchar`. And R's `as.character(double)` is not
`%.15g`; the C renderer declines in the one band it cannot reproduce (0 wrong in
34,099 doubles, 0 declines on real model literals).

### 5. Chunking now triggers on model size, not line count

Chunking amortizes the parse, and parse cost tracks characters, so a model with
many short lines was being chunked when chunking made it slower:

| model | whole | chunked | |
|---|---|---|---|
| 70 KB | 0.39 s | 2.00 s | chunking **5.1x worse** |
| 636 KB | 5.42 s | 5.61 s | 1.03 |
| 1.39 MB | 15.39 s | 4.42 s | 0.29 |
| 3.14 MB | 59.75 s | 6.70 s | chunking **8.9x better** |

The knee is near 1 MB. The threshold is `options(rxode2.optExprChunkChars = )`,
default 512 KB -- set below the knee because the payoff is asymmetric. A model
between 40 lines and the size threshold is now optimized whole, so it gets
better sharing and no `rx_expr_c<i>_` temporaries. `chunkLines = 0` still forces
a single whole-model pass, and each chunk runs through the C pass either way.

### 6. `options(rxode2.compile.O=)` now reaches the compiler

Found while pinning down what optimization level the numbers above were taken
at. The level was written into the generated model's `PKG_CFLAGS`, and
`R CMD SHLIB` builds
`ALL_CFLAGS = $(PKG_CFLAGS) $(CPICFLAGS) $(SHLIB_CFLAGS) $(CFLAGS)`, so R's own
`CFLAGS` came last and the compiler took its `-O2`. The option had no effect --
every model was built at `-O2` whatever it was set to. (rxode2's own
`src/Makevars` `-O3` is overridden the same way; that one stays, since a package
must not override R's `CFLAGS`.)

Setting `CFLAGS` in the generated package Makevars does not help either, because
the user Makevars is read after it. The user Makevars is the one place whose
`CFLAGS` wins, so the model build points `R_MAKEVARS_USER` at a temporary one
for the duration of the `R CMD SHLIB` call and restores it on exit. Only the
`-O` is displaced -- the value is rebuilt from `R CMD config CFLAGS` with the
`-O` swapped, so R's hardening flags, `-g` and prefix maps carry over -- and the
real user Makevars is inlined after our line, so anyone setting `CFLAGS` there
still wins.

Verified end to end by reading each compiled model's DWARF producer string, not
by trusting the option. `-O3` buys nothing over `-O2` on the MM fit, but making
the option work is what allowed the `-O3` row in the table above to be measured
at all.

## Where `rxOptExpr()` pays for itself

The pass has a **one-off** cost and a **per-iteration** saving, so whether it is
worth running is a question of how many iterations follow -- and the answer
turns out to depend strongly on the model.

### A realistic model: 2-compartment oral Michaelis-Menten

`two.compartment.oral.MM.model` from `nlmixr2test/cpt` (`Oral_2CPTMM.csv`, the
`U070` case): 3 states, 6 etas, nonlinear elimination, 7920 records, 120
subjects. FOCEi run to convergence, cold, median of 2 -- **283 iterations and
objective 69544.56 in every run**, so the arms are like-for-like.

**Everything below is at `-O2`.** Both rxode2 and the generated model ask for
`-O3` -- the package through `PKG_CFLAGS`, the model through
`getOption("rxode2.compile.O", "3")`, which `R/rxode2.R:2269` writes into a
Makevars as `PKG_CFLAGS=-O3` -- and in both cases R overrides it. R's
`ALL_CFLAGS = $(PKG_CFLAGS) $(CPICFLAGS) $(SHLIB_CFLAGS) $(CFLAGS)` puts the
package flags *first* and R's own `CFLAGS` (which carries `-O2`) last, and gcc
takes the last `-O` it is given. `R CMD SHLIB` on rxode2's generated Makevars
emits `-O3 ... -O2`, so the model is built at `-O2`.

That matters because the old argument for dropping the pass was that the C
compiler already does this job. At `-O2` it demonstrably does not: the MM model
gains 7.5% per iteration from a pass gcc had every opportunity to perform
itself.

| | `optExpression=TRUE` | `FALSE` |
|---|---|---|
| per iteration | **0.1914 s** | 0.2070 s |
| `$time` optimize | 54.16 s | 58.57 s |
| total | **56.61 s** | 60.66 s |

**7.5% faster per iteration**, 6.7% faster end to end. Capped at 30 iterations
instead (3 reps, same objective and iteration count) it is 5.3% per iteration,
and there the reps do not overlap at all -- 0.2502/0.2502/0.2520 s against
0.2638/0.2653/0.2704 s.

The one-off, timed directly on the models FOCEi optimizes:

| | lines | size | C | R |
|---|---|---|---|---|
| `inner` | 45 | 6.8 KB | 0.066 s | 0.221 s |
| `predOnly` | 39 | 1.2 KB | 0.025 s | 0.041 s |
| total | | | **0.091 s** | 0.262 s |

At 0.0156 s saved per iteration that is a **break-even of 6 iterations** with the
C pass, or 17 with the R one -- against a fit that takes 283. Measured a second
way, as the difference in `$time` setup pooled over 10 cold builds, the gap is
+0.050 s; but the spread *within* one arm is 0.711 s because setup includes the
C compile, so that route is noise-dominated and the direct timing above is the
one to trust.

### A linear model, for contrast: it does not pay off there

The same experiment on a 3-compartment oral PK model with lag and F, 9 etas, 40
subjects -- linear elimination -- run to convergence with identical objective and
identical 570 evaluations:

| | `optExpression=TRUE` | `FALSE` |
|---|---|---|
| median fit | 137.40 s | 135.60 s |

There the pass is ~1.3% **slower** end to end and break-even sits above 570
evaluations. Isolated per-evaluation timings on that model were 1-11% depending
on `-O`, but at ~17 ms over 5 reps they could not separate 1% from 10%.

The difference between the two is what the CSE has to work with. Michaelis-Menten
elimination puts `(VM*centr/Vc)/(KM+centr/Vc)` in the right-hand side, and its
sensitivity expansion repeats that subexpression across every eta; a linear model
generates far less redundancy for the pass to remove. So the honest summary is
that `rxOptExpr()` pays for itself quickly on the nonlinear models it was meant
for, and roughly breaks even on linear ones -- not that it is a uniform win.

Two caveats worth stating:

- My earlier claim that the CSE "is not more helpful at `-O0`" is **withdrawn**:
  `rxode2.compile.O` is inert for the reason above, so every arm of that sweep
  was `-O2` whatever the option said. (It is also only consulted when the model
  *text* changes, since the model cache keys on text -- a second reason that
  sweep measured nothing.)

  Tested properly, by making the option effective and confirming the level from
  each model's DWARF producer string rather than from the option:

  | | `-O2` | `-O3` |
  |---|---|---|
  | per iteration, `optExpression=TRUE` | 0.2302 s | 0.2299 s |
  | per iteration, `optExpression=FALSE` | 0.2453 s | 0.2426 s |
  | CSE gain | **6.1%** | **5.2%** |

  `-O3` buys nothing here -- the two `TRUE` arms overlap completely across reps
  (0.2268-0.2378 against 0.2291-0.2304) -- but the CSE still pays **5.2% at
  `-O3`**, where the `TRUE` and `FALSE` reps do not overlap at all. So gcc does
  not perform this elimination itself even when asked to optimize fully, and the
  "compilers took over the job" explanation does not hold.

**What this PR changes** is the one-off side: the pass itself drops from 0.262 s
to 0.091 s on the MM model, and from ~2.9 s to ~0.04 s on a wbc-shaped model, so
what has to be repaid shrinks and break-even moves down with it. The
per-iteration side is untouched -- the optimized text is byte-identical.

## FOCEi and SAEM, branch against `main`

`origin/main` and the branch each installed into their **own** library,
nlmixr2est run against each, cold (`rxClean()` + `rxForget()`):

| | FOCEi | SAEM |
|---|---|---|
| objective | 133.75587932 both | Inf both (control settings, not the change) |
| theta | identical to 10 dp | identical to 10 dp |
| omega | identical to 10 dp | identical to 10 dp |
| cold elapsed, median of 3 | 4.66 s -> **3.48 s** | 20.92 s -> **19.73 s** |

FOCEi is the real exercise: it runs `rxOptExpr()` on four models where saem runs
one, and it is 25% faster cold. These are short fits dominated by setup, which
is exactly the part this PR moves to C; they say nothing about the
per-evaluation question above.

Two methodology traps this ran into, recorded because both produced confident
wrong numbers:

- The first A/B compared against an rxode2 that had been installed **from this
  branch**, so both arms ran identical code and "identical results" proved
  nothing.
- `R CMD INSTALL .` **reuses the `.o` files already in `src/`**, and `load_all`
  had built those for debugging. That library had a debug solver and showed a
  68% slower `rxSolve` and 43% slower saem. Both vanished after
  `rm -f src/*.o src/*.so`; the solve is 0.0280 s either way.

## Behavior change

One degenerate form no longer parses: an `if`/`while` with no body at all and
no braces (`if (a > 1)`, `while (a > 1)`, `if (a > 1) else b <- 1`). This cannot
be kept alongside the fix -- allowing an empty body is exactly what makes
`if (a > 1) b <- 1` ambiguous between the body and the next statement. Write
`{}` or `;`; both parse as before, and a block containing only a comment is
unaffected. Documented in NEWS and tested.

Model text that is only whitespace and comments is now a blank model, as an
empty string always was.

## Verification

- Full suite: **39713 pass, 0 fail, 0 error**, with every model built at `-O3`.
- Two differential fixtures, captured from the R implementations in fresh
  subprocesses and replayed as tests: 576 `rxFromSE` + 139 `rxToSE` pairs, and
  19 `rxOptExpr` pairs (`chunkLines = 0L`, the whole-model reference).
- `tests/testthat/test-optexpr-cse-c.R` pins the C common-subexpression pass:
  byte-exactness against the R implementation on a second-order model,
  output independence from the thread count, and the negative-literal and
  nested-replacement cases that earlier versions got wrong.
- dparser re-entrancy proven before any threading was written --
  `inst/tools/dparserReentrancy.c`, 48000 concurrent parses, clean under TSAN
  and ASAN.
- Reviewed with Antigravity (Gemini) over four rounds. It found the `malloc`
  blocks that an OOM `longjmp` out of `Rf_mkChar()` would strand -- fixed with
  `R_alloc` and `R_UnwindProtect` -- and that the grammar edge cases were
  untested, which is where `test-parse-empty-statement.R` came from. Its final
  round claimed a parse-tree leak in `src/seParse.h`; that one is wrong --
  48000 translations grow RSS by 92 kB, so `free_D_Parser()` reclaims the tree.

## Follow-up, not in this PR

Pruning in C. `.rxPrune()` is a recursive R AST walk that re-parses the emitted
true-branch text at every `if`/`else` (`R/rxPrune.R:79`), and it would reuse
most of the machinery added here. It cannot be fused with the CSE counting pass,
though: pruning must run *before* symengine (symengine cannot take `if`/`else`)
and the CSE must run *after* it, both because symengine inlines CSE temporaries
straight back out and because the redundant text is the text symengine itself
generates. It is also not per-statement independent -- `envir$.def1` is a prefix
fact that changes the emitted line -- so it needs one sequential scan before the
rest can be parallelized.

A model of ~32000 statements fails to parse at all. That is a parser limit
worth its own look, unrelated to the ambiguity fixed here.


mattfidler:
@hiddevandebeek this is better than the chunking option so I removed it.  If you are still seeing issues we could possibly bring it back, though since it is already parallelized now I don't  think it is worth it.

As a note, this will have a proper common sub expression optimization instead of the chunking with a partial subexpression optimization.

Also, I measured again and it still is worthwhile on long running models (like I did years ago).  It is even more worthwhile now that it  is much much faster.

PR #1319, opened 2026-08-31
fix(ui): promote an ini() comment only where the statement is complete

### LINKED ISSUE #1318: A comment inside an unclosed call in `ini({})` is replaced by `;`, breaking the re-parse
`.rxReplaceCommentWithLabel()` rewrites trailing comments in `ini({})` into
`label()` calls by substituting the comment with `;` and re-parsing. When the
comment sits on a line where the R expression is **not yet syntactically
complete** — i.e. between an opening `(` and its closing `)` — the substitution
injects a bare `;` into the middle of the call and the re-parse fails.

### Reproducible example

```r
options(keep.source = TRUE)   # essential; see "Why this hides in CI" below
library(rxode2)

src <- '
f <- function() {
  ini({
    lka <- 0.5
    lcl <- c(
      1.0   # <- comment on a line where the expression is NOT yet complete
    )
    propSd <- 0.3
  })
  model({
    ka <- exp(lka)
    cl <- exp(lcl)
    d/dt(central) <- -cl / 3 * central
    Cc <- central / 3
    Cc ~ prop(propSd)
  })
}'

fn <- file.path(tempdir(), "m.R"); writeLines(src, fn)
env <- new.env(); source(fn, local = env, keep.source = TRUE)
rxode2::rxode2(env$f)
```

### Actual

```
Error in parse(text = paste(.ret, collapse = "\n"), keep.source = FALSE) :
  <text>:5:13: unexpected ';'
4:     lcl <- c(
5:       1.0   ;
               ^
Calls: <Anonymous> ... .rxReplaceCommentWithLabel -> deparse -> eval -> parse
```

The comment has been replaced by a bare `;` inside the unclosed `c(`.
`x <- 1 ; label("...")` is valid, so the substitution is fine on a *complete*
line; it is only invalid while the expression is still open.

### Expected

The model builds, as it does when the same comment is moved to a line where the
expression is complete.

### What is and is not the trigger

Comment position relative to the unclosed call is the only thing that matters.
Number of `#` on the line, and whether the comment text looks like code, are
both irrelevant.

| Variant | Result |
|---|---|
| No comment anywhere in the multi-line call | PASS |
| Single `#`, plain text, inside the unclosed call | **FAIL** |
| Single `#`, code-like text (`# omega^2 = log(0.3^2 + 1)`), inside | **FAIL** |
| Two `#` on one line, plain text between, inside | **FAIL** |
| Two `#` on one line, code-like text between, inside | **FAIL** |
| Comment on the line that *opens* the call (`~ c(  # note`) | **FAIL** |
| Comment on the *closing-paren* line (`)  # note`) | PASS |
| Same call written on a single line, with trailing comment | PASS |
| Two `#` on an ordinary complete `ini()` line | PASS |
| Blank line (no comment) inside the call | PASS |

Not specific to correlated-eta `~ c(...)` blocks: a plain `lcl <- c(\n 1.0 # cmt\n)`
fails identically, and a single-eta `etaX ~ c(\n 0.09 # var\n)` fails too.

### Why this hides in CI

Reproduction requires `keep.source = TRUE`:

```
keep.source = TRUE   ->  FAIL
keep.source = FALSE  ->  PASS
```

`devtools::load_all()` keeps srcrefs, so this fires during development, whereas
a normally-installed package does not, so the same model builds under pkgdown /
`R CMD check` on an installed tree. In nlmixr2lib this shows up as
`Ahmed_2016_lorenzosOil` failing `checkModelConventions()` and its vignette
render locally, while the identical commit renders that vignette successfully in
CI.

### Suggested fix

Only substitute the comment when the accumulated expression is complete — e.g.
track paren/bracket depth while walking the srcref lines and leave comments
inside an open expression alone (or drop them) rather than emitting `;`.

### Session info

- rxode2 5.1.7
- R 4.6.1, x86_64-pc-linux-gnu

billdenney:
Fixes #1318.

`.rxReplaceCommentWithLabel()` rewrote a trailing `#` comment in `ini({})` into
`; label("...")` on any line carrying one, regardless of whether the R statement on
that line had finished. A comment between an opening `(` and its `)` therefore put
the `;` in the middle of the statement and the re-parse failed with a bare
`unexpected ';'` pointing into regenerated text rather than at the source line.

The same defect breaks a line that is incomplete because it *ends in an operator*,
which bracket-depth tracking would not catch:

```r
ini({
  tcl <- 0.45 +   # note
    0.1
  eta.cl + eta.v ~  # note
    c(1, 0.01, 1)
})
```

So instead of counting brackets, this asks R's own parser whether the accumulated
statement is complete before promoting. A comment inside an unfinished statement now
stays a comment and is dropped; a comment on the line that *closes* the statement
still becomes that parameter's label, unchanged.

### Second, silent defect fixed here too

`.regCommentOnBlankLine` allowed leading spaces only, so a **tab**-indented
comment-only line fell through to the label branch, whose code group then captured
just the tab. The bare `; label("...")` that produced parses -- a leading `;` is
legal -- so there was no error: the comment silently became the label of the
**preceding** parameter. On 5.1.7 this source gives `tka` the label
`"tab indented comment"`:

```r
ini({
  tka <- 0.45
	# tab indented comment
  tcl <- 1
})
```

### Why it hides in CI

The promotion only runs with source refs intact, so `devtools::load_all()` hits it
while an installed package does not. In nlmixr2lib this showed as
`Ahmed_2016_lorenzosOil` failing `checkModelConventions()` and its vignette locally
while the identical commit rendered green in CI.

### Tests

23 assertions in `tests/testthat/test-ui.R` covering all eight statement shapes, both
tab cases, and exact generated-source assertions. The tab regression guard builds the
model and asserts `iniDf$label` is all-`NA`, pinning the silent misattribution.

Full suite: `FAIL 0 | WARN 33 | SKIP 7 | PASS 40109`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -990,6 +990,26 @@ mod |> ini(prior(eta.cl, eta.v) ~ invWishart(4))
   dropped silently -- the model still parsed and built, it simply lost the label
   (#1205).
 
+- A `#` comment inside an `ini({})` statement that spans more than one line no
+  longer breaks the model.  The comment was promoted by appending
+  `; label("...")` whether or not the statement on that line had finished, so a
+  comment between an opening `(` and its `)` -- or on a line ending in an
+  operator such as `+` or `~` -- put the `;` in the middle of the statement and
+  the model failed with a bare `unexpected ';'` pointing into regenerated text
+  rather than at the offending source line.  A comment is now promoted only
+  where it trails a complete statement; one inside an unfinished statement stays
+  a comment and is dropped.  A comment on the line that closes the statement
+  still becomes that parameter's label.  As with #1195 the promotion only runs
+  when source refs are kept, so the same model built fine from an installed
+  package while failing under `keep.source = TRUE` (#1318).
+
+- A comment-only `ini({})` line indented with a tab is no longer turned into the
+  label of the preceding parameter.  The comment-only test allowed leading
+  spaces only, so a tab-indented comment fell through to the label branch, whose
+  code portion then captured just the tab.  The bare `; label("...")` that
+  produced parses -- a leading `;` is legal -- so there was no error and the
+  comment silently became the label of the parameter above it (#1318).
+
 ### Estimation / symengine translation
 
 - A model using the modulo operator `%%` can now be estimated.  `%%` was
--- a/R/ui.R
+++ b/R/ui.R
@@ -1,3 +1,18 @@
+#' Do the accumulated `ini({})` lines parse as a complete statement?
+#'
+#' A `; label()` may only be appended where the statement is finished; the lines
+#' are wrapped in a block so completeness is judged in the context they occupy.
+#'
+#' @param lines `ini({})` source lines accumulated for the current statement
+#' @return `TRUE` when the lines parse as complete R code
+#' @author Matthew Fidler
+#' @noRd
+.rxIniLinesComplete <- function(lines) {
+  .txt <- paste0("{\n", paste(lines, collapse="\n"), "\n}")
+  !inherits(try(parse(text=.txt, keep.source=FALSE), silent=TRUE),
+            "try-error")
+}
+
 #' Replace comments with label()
 #'
 #' @param src source function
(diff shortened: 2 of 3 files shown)


mattfidler:
Thanks

billdenney:
This also fixes #259

PR #1320, opened 2026-08-31
fix(tad): count infusions in the dose history when the first dose is a bolus

### LINKED ISSUE #1316: `tad()` and `dosenum()` silently skip every infusion when the subject's first dose is a bolus
## Summary

An infusion updates the dose history (`tad()`, `dosenum()`, `tlast()`) correctly
as long as the subject's dosing record starts with an infusion. When the
subject's first dose is a bolus, every infusion for that subject is silently
skipped: `dosenum()` does not increment and `tad()` keeps counting from the
earlier dose. Boluses are still counted, and once the first dose is an
infusion, later boluses do no harm. No warning, no message.

The two event tables in the example below differ only in the `rate` of the
**first** dose. The second dose is identical in both, but is counted in one and
not in the other.

Solving itself is unaffected: the infusion is delivered on schedule and the
amounts are right. What breaks is the dose-history bookkeeping, so TAD-based
diagnostics (`vpcPlotTad()`, anything binned on `tad`) are wrong.

## Reproducible example

```r
library(rxode2)  # 5.1.6

m <- rxode2({
  ka <- 1; cl <- 5; v <- 50
  d/dt(depot)  <- -ka * depot
  d/dt(center) <-  ka * depot - (cl / v) * center
  tadx <- tad()
  dn   <- dosenum()
})

# two event tables, differing ONLY in the rate of the first dose
ev <- function(rate1) {
  e <- rbind(
    data.frame(time = 0,   evid = 1, amt = 100, cmt = "depot",  rate = rate1),
    data.frame(time = 240, evid = 1, amt = 50,  cmt = "center", rate = 40),
    data.frame(time = c(241, 250), evid = 0, amt = 0, cmt = "center", rate = 0))
  e$id <- 1
  e
}

f <- function(rate1) rxSolve(m, ev(rate1), returnType = "data.frame")[, c("time", "tadx", "dn")]

f(40)   # first dose is an infusion
#>  time tadx dn
#>   241    1  2
#>   250   10  2

f(0)    # first dose is a bolus -- second dose silently skipped
#>  time tadx dn
#>   241  241  1
#>   250  250  1
```

In the second call the dose at t = 240 leaves `dosenum()` at 1 and `tad()`
running from t = 0.

Varying the dosing sequence pins the rule down. In the table, `bolus(t)` and
`INF(t)` are a bolus or infusion dosed at time `t` (boluses into `depot`,
infusions into `center`, unless noted otherwise). An infusion is skipped if and
only if the subject's first dose was a bolus — equivalently, if and only if a
bolus precedes the subject's first infusion:

| dosing sequence | infusion updates `tad`/`dosenum`? |
|---|---|
| INF only | yes |
| INF(0) → INF(240) | yes |
| INF(0) → bolus(120) → INF(240) | yes |
| INF(0) → bolus(240) | yes |
| bolus(0) → INF(240) | **no** |
| bolus(0) → bolus(120) → INF(240) | **no** |
| bolus(0) → INF(120) → INF(240) | **no** (both infusions skipped) |
| bolus(0, center) → INF(240, center) | **no** (same compartment) |
| bolus(0) → INF(120) → bolus(240) | **no** (the infusion alone is skipped) |

It is not specific to `rate`. Giving the infusion a `dur` column instead
reproduces it exactly:

```r
ev2 <- function(dur1) {
  e <- rbind(
    data.frame(time = 0,   evid = 1, amt = 100, cmt = "depot",  dur = dur1),
    data.frame(time = 240, evid = 1, amt = 50,  cmt = "center", dur = 1.25),
    data.frame(time = c(241, 250), evid = 0, amt = 0, cmt = "center", dur = 0))
  e$id <- 1
  e
}

rxSolve(m, ev2(2.5), returnType = "data.frame")[, c("time", "tadx", "dn")]
#>  time tadx dn
#>   241    1  2
#>   250   10  2

rxSolve(m, ev2(0), returnType = "data.frame")[, c("time", "tadx", "dn")]
#>  time tadx dn
#>   241  241  1
#>   250  250  1
```

That the solve is unaffected can be seen by observing across the infusion —
`center` goes from ~0 at t = 239 to 47.0 at t = 241.25, i.e. the 50 units are
delivered over 1.25 h as asked, while `dn` stays at 1 throughout.

## Cause

`inst/include/rxode2parseHandleEvid.h`, `handleTlastInline()`:

```c
double curDose = getDoseIndex(ind, ind->idx), tinf = NA_REAL;   // line 146
if (handleTlastInlineUpateDosingInformation(ind, &curDose, &tinf) == 0) return;
ind->dosenum++;
ind->tlast = _time;
```

`curDose` is read with `ind->idx`, the current record. But the infusion branch of
`handleTlastInlineUpateDosingInformation()` switches index, and asks for the
duration with `ind->ixds`, the dose counter:

```c
case EVIDF_INF_DUR:                                              // line 106
case EVIDF_INF_RATE:
  if (curDose[0] <= 0) {
    return 0;
  } else {
    if (ind->fns->getdur) tinf[0] = ind->fns->getdur(ind->ixds, ind, 2, &p);
    else tinf[0] = _getDur(ind->ixds, ind, 2, &p);                // line 113
    if (!ISNA(tinf[0])) {
      curDose[0] = tinf[0] * curDose[0];
      return 1;
    } else {
      return 0;                                                   // silent exit
    }
  }
```

`_getDur()` (`src/handle_evid.cpp`) reads the amount at `l = ind->ixds` and
searches forward for the matching infusion-off record, the one carrying `-dose`.
When `ixds` is not pointing at this infusion's own record the search cannot
succeed, and with `backward == 2` the miss returns `NA_REAL` instead of raising:

```c
p[0] = l+1;
while (p[0] < ind->ndoses && getDoseNumber(ind, p[0]) != -dose){
  p[0]++;
}
if (getDoseNumber(ind, p[0]) != -dose){
  if (backward==2) return(NA_REAL);                              // line 84
  ...
  (Rf_errorcall)(R_NilValue, "infusion end cannot be found");
}
```

`NA` then makes `handleTlastInlineUpateDosingInformation()` return 0, and
`handleTlastInline()` returns before `ind->dosenum++` and `ind->tlast = _time`
ever run. The dose is dropped from the history with nothing reported — the
`backward == 2` branch is what turns a lookup failure into silence.

These lines are unchanged on `main` as of a3cedce (2026-08-31):
`rxode2parseHandleEvid.h` lines 106-113 and 146-148, `handle_evid.cpp` line 84.

## Session info

```
R version 4.4.3 (2025-02-28)
Platform: aarch64-apple-darwin24.2.0
Running under: macOS Sequoia 15.7.7

Matrix products: default
BLAS:   /opt/homebrew/Cellar/openblas/0.3.32/lib/libopenblasp-r0.3.32.dylib 
LAPACK: /opt/homebrew/Cellar/r/4.4.3_1/lib/R/lib/libRlapack.dylib;  LAPACK version 3.12.0

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

time zone: Asia/Tokyo
tzcode source: internal

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] rxode2_5.1.6

loaded via a namespace (and not attached):
 [1] vctrs_0.7.3           cli_3.6.6             rlang_1.3.0          
 [4] generics_0.1.4        S7_0.2.2              data.table_1.17.6    
 [7] RcppParallel_5.1.11-1 glue_1.8.1            backports_1.5.1      
[10] dparser_1.3.1-13      sys_3.4.3             scales_1.4.0         
[13] grid_4.4.3            tibble_3.3.1          fastmap_1.2.0        
[16] lifecycle_1.0.5       memoise_2.0.1         compiler_4.4.3       
[19] dplyr_1.2.1           RColorBrewer_1.1-3    Rcpp_1.1.2           
[22] pkgconfig_2.0.3       farver_2.1.2          R6_2.6.1             
[25] tidyselect_1.2.1      pillar_1.11.1         PreciseSums_0.7      
[28] magrittr_2.0.5        checkmate_2.3.4       tools_4.4.3          
[31] gtable_0.3.6          lotri_1.0.4           cachem_1.1.0         
[34] ggplot2_4.0.3
```

---

*Prepared with [Claude Code](https://claude.com/claude-code). Everything above
was checked by running it: the examples and their printed output are from an
actual run on the environment listed, the dosing-sequence table is one run per
row, and the source excerpts and line numbers were read from the installed
package and from the files at the commit cited rather than recalled.*

mattfidler:
Fixes #1316.

## The bug

`tad()`, `tafd()`, `tlast()`, `dosenum()` and `dose()` silently skipped an
infusion whenever a subject's dose record started with a bolus. The infusion was
delivered on schedule -- only the dose-history bookkeeping broke, so TAD-based
diagnostics (`vpcPlotTad()`, anything binned on `tad`) were wrong with nothing
reported.

## Cause

`handleTlastInlineUpateDosingInformation()` looked the infusion duration up with
`ind->ixds`, the solver's running dose counter. `handleEvid1()` syncs that
counter via `syncIdx()` before calling `handle_evid()`, but the output pass in
`src/rxode2_df.cpp` calls `handleTlastInline()` directly with only `ind->idx`
set and never advances or syncs `ixds`, so there it pointed at an unrelated
dose. The lookup then either

- failed, and `_getDur(..., backward=2)` returned `NA` -- the dose was dropped
  from the history entirely, leaving `dosenum()` un-incremented and `tad()`
  counting from the earlier dose; or
- silently returned a *different* infusion's duration, which made `dose()`
  report the wrong amount. This one bit even the sequences the issue lists as
  working: with two rate-40 infusions of different durations, the second
  reported the first one's amount.

## Fix

Derive the dose index from the record being handled (`ind->ix[ind->idx]`)
instead of trusting `ixds`. The new `handleTlastInlineDoseIndex()` tries the
already-synced `ixds` first, then the existing `getDoseNumberFromIndex()`
bisection, then a linear scan, and only falls back to `ixds` for extra doses
(`idx < 0`), whose amount lives in `extraDoseDose` rather than `idose` and for
which there is no dose index to recover -- unchanged behavior there.

`getDoseNumberFromIndex()` moved earlier in the header so it is defined before
use; it is otherwise untouched.

## Also in here

Two pre-existing problems an Antigravity review turned up while checking this
change:

- `_getDur()` read `ind->idose[l]` through `getDoseNumber()` *before* either
  branch's bounds check, so a dose counter run past the last dose (which the
  `syncIdx()` comment already documents as possible) produced an out-of-bounds
  read ahead of the check meant to catch it. A single `l < 0 || l >= ndoses`
  guard is now hoisted to the top, keeping the `NA` return for `backward == 2`
  and both existing messages; the two now-subsumed inner checks are gone.
- `getDoseNumberFromIndex()`'s bisection assumes `idose` stays ascending in
  record index, but `_rxPushDose()` re-sorts the unprocessed `idose` suffix by
  *time*, so it can miss a dose that is present. Hence the linear scan before
  the fallback.

## Verification

`tests/testthat/test-tad-infusion-after-bolus.R` covers the reported case, a
modeled-duration infusion, two infusions after a bolus, a bolus into the infused
compartment, `dose()` reporting each infusion's own amount, `addl`-expanded
boluses, a reset (`evid=3`) in between, and a modeled `alag` that reorders the
records so `ind->ix` is not the identity permutation. Every one of them fails on
the unfixed build.

All nine dosing sequences from the issue's table were re-run and now update the
history. Locally, a targeted subset of 14,175 tests and a 7,792-test subset after
the hardening both pass with no failures; the full suite is left to CI.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -684,6 +684,16 @@
 
 ## Bug fixes
 
+- `tad()`, `tafd()`, `tlast()`, `dosenum()` and `dose()` no longer skip an
+  infusion when the subject's dosing record starts with a bolus.  The dose
+  history asked for the infusion duration with the solver's running dose
+  counter (`ind->ixds`), which the output pass never advances, so the lookup
+  either failed -- the infusion was then dropped from the history entirely,
+  leaving `dosenum()` un-incremented and `tad()` counting from the earlier
+  dose -- or silently returned a different infusion's duration, which made
+  `dose()` report the wrong amount.  The duration is now looked up from the
+  record being handled.  Solving itself was never affected (#1316).
+
 - Parsing a model no longer corrupts the caller's `PROTECT` stack.  The
   translation table and `_goodFuns` were claimed on the protect stack by one
   function and released by another, with the whole parse in between; an
--- a/inst/include/rxode2parseHandleEvid.h
+++ b/inst/include/rxode2parseHandleEvid.h
@@ -107,7 +107,9 @@ static inline int getDoseNumberFromIndex(rx_solving_options_ind *ind, int idx) {
 // (rxode2_df.cpp) calls handleTlastInline() without advancing or syncing it, so
 // ixds can point at an unrelated dose there.  Recover the index from ind->idx
 // whenever the record is a real (non-extra) dose, and only fall back to ixds
-// for extra doses, whose idx does not map into idose.
+// for extra doses (idx < 0).  An extra dose's amount lives in extraDoseDose,
+// not in idose, so there is no dose index to recover for it; it keeps the
+// long-standing ixds behavior.
 static inline int handleTlastInlineDoseIndex(rx_solving_options_ind *ind) {
   if (ind->idx < 0) return ind->ixds;
   int cur = ind->ix[ind->idx];
--- a/inst/include/rxode2parseHandleEvid.h
+++ b/inst/include/rxode2parseHandleEvid.h
@@ -110,7 +110,12 @@ static inline int getDoseNumberFromIndex(rx_solving_options_ind *ind, int idx) {
 // for extra doses, whose idx does not map into idose.
 static inline int handleTlastInlineDoseIndex(rx_solving_options_ind *ind) {
   if (ind->idx < 0) return ind->ixds;
-  int m = getDoseNumberFromIndex(ind, ind->ix[ind->idx]);
+  int cur = ind->ix[ind->idx];
+  // the solving pass syncs ixds before handle_evid(), so it usually already
+  // matches and the bisection can be skipped
+  if (ind->ixds >= 0 && ind->ixds < ind->ndoses &&
+      ind->idose[ind->ixds] == cur) return ind->ixds;
+  int m = getDoseNumberFromIndex(ind, cur);
   return (m == -1) ? ind->ixds : m;
 }
 
--- a/inst/include/rxode2parseHandleEvid.h
+++ b/inst/include/rxode2parseHandleEvid.h
@@ -116,7 +116,15 @@ static inline int handleTlastInlineDoseIndex(rx_solving_options_ind *ind) {
   if (ind->ixds >= 0 && ind->ixds < ind->ndoses &&
       ind->idose[ind->ixds] == cur) return ind->ixds;
   int m = getDoseNumberFromIndex(ind, cur);
-  return (m == -1) ? ind->ixds : m;
+  if (m != -1) return m;
+  // _rxPushDose() appends doses at the end of all_times and then re-sorts the
+  // unprocessed idose suffix by time, so idose is not guaranteed to stay
+  // ascending in the record index and the bisection can miss a dose that is
+  // there.  Scan for it before giving up.
+  for (int j = 0; j < ind->ndoses; j++) {
+    if (ind->idose[j] == cur) return j;
+  }
+  return ind->ixds;
 }
 
 static inline int handleTlastInlineUpateDosingInformation(rx_solving_options_ind *ind, double *curDose, double *tinf) {
--- a/inst/include/rxode2parseHandleEvid.h
+++ b/inst/include/rxode2parseHandleEvid.h
@@ -89,6 +89,31 @@ static inline void handleInfusionGetEndOfInfusionIndex(int idx, int *infEixds,
 	}
 }
 
+static inline int getDoseNumberFromIndex(rx_solving_options_ind *ind, int idx) {
+  // bisection https://en.wikipedia.org/wiki/Binary_search_algorithm
+  int l = 0, r = ind->ndoses-1, m=0, idose = 0;
+  while(l <= r){
+    m = FLOOR((l+r)/2);
+    idose= ind->idose[m];
+    if (idose < idx) l = m+1;
+    else if (idose > idx) r = m-1;
+    else return m;
+  }
+  return -1;
+}
+
+// Dose index (into ind->idose) of the record currently being handled.
+// ind->ixds is the solver's running dose counter; the output/lhs pass
+// (rxode2_df.cpp) calls handleTlastInline() without advancing or syncing it, so
+// ixds can point at an unrelated dose there.  Recover the index from ind->idx
+// whenever the record is a real (non-extra) dose, and only fall back to ixds
+// for extra doses, whose idx does not map into idose.
+static inline int handleTlastInlineDoseIndex(rx_solving_options_ind *ind) {
+  if (ind->idx < 0) return ind->ixds;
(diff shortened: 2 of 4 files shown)


mattfidler:
Follow-up filed as #1321 for the extra-dose (`ind->idx < 0`) duration lookup, which this PR deliberately leaves on its long-standing behavior.

mattfidler:
Antigravity round 4 confirmed the forward-scan short-circuit in `c30ebde2f` fixes the cross-subject read without changing any correct answer, and found nothing further in this PR's changes. Its three other findings are all pre-existing and untouched here; filed as #1322.

PR #1326, opened 2026-09-01
perf: coerce covariate columns to double once, not once per row

### LINKED ISSUE #1325: Logical/integer covariate columns are up to 90x slower than double and scale superlinearly
When a data column referenced as a covariate in the model block is stored as `logical` or `integer` rather than `double`, `rxSolve()` is dramatically slower and its cost becomes superlinear in the number of rows. Results are numerically identical, so this is purely a performance issue — but a severe one on large simulations.

With doubles, throughput is roughly flat in row count (as expected). With logical or integer, throughput collapses as rows grow:

| path | type | 20k rows | 40k | 80k | 160k | scaling exponent |
|---|---|---|---|---|---|---|
| `rxode2::rxSolve` | double  | 0.09 s | 0.21 s | 0.43 s | **1.10 s** | 1.19 |
| `rxode2::rxSolve` | integer | 1.65 s | 5.32 s | 21.37 s | **81.46 s** | 1.89 |
| `rxode2::rxSolve` | logical | 1.90 s | 5.06 s | 18.27 s | **100.90 s** | 1.90 |
| nlmixr2 `est="rxSolve"` | double  | 0.81 s | 1.22 s | 2.51 s | 8.17 s | 1.10 |
| nlmixr2 `est="rxSolve"` | integer | 2.36 s | 6.25 s | 27.00 s | 75.46 s | 1.71 |
| nlmixr2 `est="rxSolve"` | logical | 1.75 s | 3.95 s | 16.45 s | 58.57 s | 1.73 |

At 160k rows that is a **92x** penalty through `rxSolve()` directly. It reproduces without nlmixr2 in the call path, so it looks like it lives in rxode2's data handling rather than in nlmixr2est.

Note that `ID` and `TIME` are `integer` in every one of these runs — only the model-referenced covariate `flag` changes storage mode. So it appears specific to covariate columns, not to event-table columns generally.

## Reproducible example

```r
library(rxode2)

mk <- function(n_row, type) {
  flag <- rep(c(0, 1), length.out = n_row)
  data.frame(
    ID = rep(seq_len(n_row/50), each = 50),
    TIME = rep(seq_len(50), times = n_row/50),
    x = seq(-100, 0, length.out = n_row),
    flag = switch(type,
                  double = as.double(flag),
                  integer = as.integer(flag),
                  logical = as.logical(flag)),
    EVID = 0, AMT = 0
  )
}

m <- rxode2({
  slope <- slope_a*(1 - flag) + slope_b*flag
  eff <- slope*x
})
pars <- c(slope_a = -0.90, slope_b = -1.29)

solve_it <- function(d) rxSolve(m, params = pars, events = d, returnType = "data.frame")
invisible(solve_it(mk(20000, "double")))  # warm up

for (ty in c("double", "integer", "logical")) {
  for (n in c(20000, 40000, 80000, 160000)) {
    d <- mk(n, ty)
    s <- system.time(invisible(solve_it(d)))[["elapsed"]]
    cat(sprintf("%-8s %7d rows %8.2f s (%7.0f rows/s)\n", ty, n, s, n/s))
  }
}

# identical results
all.equal(solve_it(mk(20000, "double"))$eff, solve_it(mk(20000, "logical"))$eff)  # TRUE
all.equal(solve_it(mk(20000, "double"))$eff, solve_it(mk(20000, "integer"))$eff)  # TRUE
```

## Why it matters

Logical covariates arise naturally from ordinary R code — `flag = x == "value"`, or a `data.frame(is_a = c(FALSE, TRUE, FALSE))` crossed into a simulation grid — so this is easy to hit without realising it. In our model-based meta-analysis a 4-million-row simulation took ~60 minutes; converting two logical indicator columns to numeric (identical values, identical results) brought the same simulation to ~8 minutes.

The workaround is simply `as.numeric()` on any column the model block reads, but the failure mode is silent: correct answers, quietly 10-90x slower.

## Environment

```
rxode2 5.1.7, nlmixr2est 7.0.3, nlmixr2 7.0.1
R version 4.6.1 (2026-06-24 ucrt), x86_64-w64-mingw32
```

billdenney:
Fixes #1325.

## The defect

`etTrans()` builds the solving data set one output row at a time, and for each
row it did this per covariate:

```cpp
nvTmp2 = as<NumericVector>(cur);   // cur = inData[covColj]
```

For a column that is already `double` that returns the same vector and costs
nothing — which is why the problem was invisible in the common case. For an
`integer` or `logical` column it calls `Rf_coerceVector`, allocating and
converting a copy of the **whole column**, on **every row**. The loop was
therefore quadratic in rows, with correct results throughout.

Each covariate column is now coerced once, before the loop, and the loop reads
it through a plain pointer. A column that is already double is still not
copied at all.

| rows | `double` | `integer` | `logical` |
|---|---|---|---|
| before, 20k | 0.12 s | 1.68 s | 2.06 s |
| before, 160k | 1.10 s | 81.5 s | 100.9 s |
| after, 160k | 1.12 s | **1.01 s** | **1.08 s** |

This matters because a logical covariate arises from ordinary R code —
`flag = x == "value"`, or a `data.frame(is_a = c(FALSE, TRUE))` crossed into a
simulation grid. The failure mode was silent: correct answers, quietly up to
90x slower.

## Second commit: the same defect in `rxSolve()`

The copy of each covariate into the solving buffer runs once per subject, and
each copy coerced its column first (`src/rxData.cpp`, three sites). Those
columns come from `etTrans()` and are almost all double — except `CMT`, which
is carried as an integer column. So a model that reads `CMT` as a covariate
re-converted the whole column for every subject: O(nsub × nrow).

Measured with 20 000 subjects / 60 000 rows, against an otherwise identical
double covariate: **3.35x before, 0.93x after**.

## Tests

Three regression tests in `tests/testthat/test-cov.R`, each verified to go
**red on the unfixed build and green on the fixed one** — a timing test that
cannot fail is worse than no test:

- `etTrans()` carries every covariate as a double whatever the input storage
  mode, the model sees identical values, and an `NA` in an integer or logical
  column resolves the same way an `NA` in a double column does (exact
  equality — these are strict).
- Integer and logical covariates stay within 4x a double covariate at
  20 000 rows (was ~14x).
- A `CMT` covariate stays within 2x an equivalent double covariate at
  20 000 subjects (was 3.35x).

The timing thresholds are loose on purpose — timing on a shared runner is
noisy — but the defects they guard against are an order of magnitude past
them, and each measurement is the minimum of three repeats so a stray GC pause
cannot fail the check.

## Verification

Full suite on Linux / R 4.6.1: **0 failures, 0 errors, 39 261 passing, 5
skipped.** No new compiler warnings.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -701,6 +701,17 @@
 
 ## Bug fixes
 
+- An `integer` or `logical` covariate column no longer makes `rxSolve()`
+  quadratic in the number of rows.  While building the solving data set each
+  covariate column was coerced to double once per output row; for a column
+  that is already double that coercion is free, but for an integer or logical
+  column it allocated and converted a copy of the whole column on every row.
+  The result was correct but progressively slower -- roughly 14x a double
+  column at 20000 rows and 90x at 160000 -- and the only hint was the timing,
+  since a logical covariate arises naturally from ordinary R code such as
+  `flag = x == "value"`.  Each covariate column is now coerced once, so every
+  storage mode solves at the speed a double column always did.
+
 - Building a model no longer hangs forever on a lock left behind by an
   interrupted session, and two processes no longer build the same model at
   once.  `rxTempDir()` exports itself with `Sys.setenv()`, so every subprocess
--- a/NEWS.md
+++ b/NEWS.md
@@ -712,6 +712,13 @@
   `flag = x == "value"`.  Each covariate column is now coerced once, so every
   storage mode solves at the speed a double column always did.
 
+- A model that reads `CMT` as a covariate no longer slows down with the number
+  of subjects.  `CMT` is carried as an integer column, and the copy of each
+  covariate into the solving buffer -- done once per subject -- coerced the
+  whole column to double every time, costing about 3x an otherwise identical
+  double covariate at 20000 subjects.  The columns are now coerced once, as
+  above.
+
 - Building a model no longer hangs forever on a lock left behind by an
   interrupted session, and two processes no longer build the same model at
   once.  `rxTempDir()` exports itself with `Sys.setenv()`, so every subprocess
--- a/src/etTran.cpp
+++ b/src/etTran.cpp
@@ -3065,6 +3065,35 @@ List etTrans(List inData, const RObject &obj, bool addCmt=false,
       stop(_("the columns that are kept must be either an underlying logical, string, factor, integer number, or real number"));
     }
   }
+  // The row loop below reads every covariate column once per output row.
+  // Coercing to double there costs a whole-column allocation and copy for each
+  // integer or logical column on each row, which makes the loop quadratic in
+  // the number of rows.  Coerce each column to double once, here, and keep the
+  // coerced columns alive in covD so covDp[] stays valid for the loop.  A
+  // column that is already double is not copied at all.
+  List covD(covCol.size());
+  std::vector<double*> covDp(covCol.size(), NULL);
+  std::vector<R_xlen_t> covDn(covCol.size(), 0);
+  for (j = 0; j < (int)(covCol.size()); j++) {
+    int covColj = covCol[j];
+    // cmt is carried as an integer compartment column, not a double covariate
+    if (hasCmt && covColj >= 0 && j == cmtI) continue;
+    SEXP curCov;
+    if (covColj >= 0) {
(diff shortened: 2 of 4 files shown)


mattfidler:
Thanks @billdenney

PR #1327, opened 2026-09-01
fix: remove code that can never run, and one result that was always discarded

billdenney:
Stacked on #<PR 1>. Base is `fix/covariate-int-logical-perf-1325`; retarget to
`main` after that merges.

Found while fixing #1325 — the covariate loop had a second row-scan whose guard
could never be true. Sweeping for the same shape (cppcheck across the C/C++
sources, `lintr::unreachable_code_linter` across `R/`) turned up 43 candidates,
of which 4 were genuine. The sweep is worth a note: the site that started this
was **not** among cppcheck's findings, because it cannot prove
`iCur+1 <= idxOutput.size()` statically; and 25 of its 36 hits were one
ABI-constrained pattern that should not be touched. The tools narrowed the
search, they did not do the triage.

## The one that changes behaviour

`populateDfdy()` translates both `_THETA_n_` and `_ETA_n_` back to their
displayed spellings, but only the THETA loop set `foundIt`. The ETA loop wrote
`ETA[n]` into the buffer and the next `if (!foundIt)` overwrote it with the raw
internal name:

```r
rxDfdy(mod)
#> "df(A)/dy(A)"  "df(A)/dy(THETA[1])"  "df(A)/dy(_ETA_1_)"
#>                                                ^^^^^^^ leaked internal name
```

Not dead code — code whose result was unconditionally discarded. Fixed by
setting `foundIt` in the ETA loop (and breaking once it matches). Nothing parses
these names back; `rxDfdy()` returns them to the user, so this only changes what
is displayed. No existing test locked the buggy form.

## The three that cannot be observed

- **`etTrans()`** searched a subject's rows for a non-NA covariate value twice:
  backwards, then forwards. The forward search was guarded by
  `idxOutput.size() < iCur+1` — never true, since `iCur < idxOutput.size()` —
  and by `iCur+1 != covCol.size()`, comparing a row index against the covariate
  count. **Repairing both guards would not help.** The enclosing loop walks the
  output rows backwards, so `addId` is set on a subject's *last* row and the
  backward search already covers every row of that subject; the forward search
  would only reach the next subject and be rejected by the `lastId` guard.
  Removed rather than repaired.
- **`rxGetErrs()`** called `free(_rxGetErrs)` inside a branch entered only when
  `_rxGetErrs` is `NULL`.
- **`.udfCallCheckNumeric()`** ended with a value expression after `stop()`.

## Tests

- `test-dfdy.R` — a real gate: **2 failures before the fix, 0 after.**
- `test-cov.R` — a *characterization* test for the `etTrans()` removal, not a
  gate. It pins the behaviour that has to survive: a subject whose covariate is
  present only in its **first** rows (`5, 5, NA, NA`) still resolves to `5`,
  which is precisely the case a forward scan would have existed to handle. It
  passes on both builds — that is the point. It proves the removal is lossless.

## Deliberately left alone

- `if (!iniSubject(...))` at ~25 solver call sites is genuinely always-false
  today (`iniSubject` has a single `return 1`), but `iniSubjectE` is published
  at slot 6 of the cross-package pointer vector and typed `int (*)(...)` in
  `inst/include/rxode2ptr.h`. Removing the checks would bake "this cannot fail"
  into code consuming a contract that says otherwise. Changing that is an ABI
  decision, not a cleanup.
- `localAbort2 == 0` / `la2 == 0` in the adjoint drivers — cppcheck does not
  model the `#pragma omp atomic write` from sibling threads.
- `j >= op->minSS` (×5), `_t < 0`, `nOrd > 0`, `omp_get_thread_num() != 0` —
  redundant-but-executing guards, or OpenMP/defensive checks where removal
  trades safety for tidiness.
- `stop()` in `switch()` defaults in `rxRaw.R`, and `with(..., return())` in
  `rxode2.R` — lintr false positives.

## Verification

Full suite on Linux / R 4.6.1: **0 failures, 0 errors, 39 269 passing, 5
skipped**, and no warnings leaked from the new tests.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -719,6 +719,13 @@
   double covariate at 20000 subjects.  The columns are now coerced once, as
   above.
 
+- `rxDfdy()` (and `rxModelVars()$dfdy`) report an ETA derivative as
+  `df(A)/dy(ETA[1])` instead of leaking the internal name
+  `df(A)/dy(_ETA_1_)`.  Both `THETA[n]` and `ETA[n]` are translated back from
+  their internal spellings for display, but the ETA translation was written
+  into the buffer and then unconditionally overwritten, so only `THETA[n]`
+  survived.
+
 - Building a model no longer hangs forever on a lock left behind by an
   interrupted session, and two processes no longer build the same model at
   once.  `rxTempDir()` exports itself with `Sys.setenv()`, so every subprocess
--- a/R/rudf.R
+++ b/R/rudf.R
@@ -516,5 +516,4 @@ rxRmFunParse <- function(name) {
   }
   stop(paste0(.udfCallFunArg(fun, args), "needs to return a length 1 numeric"),
        call.=FALSE)
-  .ret
 }
--- a/src/etTran.cpp
+++ b/src/etTran.cpp
@@ -3202,7 +3202,12 @@ List etTrans(List inData, const RObject &obj, bool addCmt=false,
                 int idxOut = i;
                 int idxIn = i;
                 double vCur = curCovD[idxInput[idxOutput[i]]];
-                // Could be NA, look for non NA value OR beginning of subject
+                // Could be NA, look for non NA value OR beginning of subject.
+                // The enclosing loop walks the output rows backwards, so addId
+                // is set on a subject's LAST row and walking iCur down from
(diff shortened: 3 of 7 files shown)


mattfidler:
Thanks @billdenney

PR #1328, opened 2026-09-01
perf(etTrans): take the output column pointers once, not once per row

billdenney:
Stacked on #1327. Base is `fix/dead-code-covariate-na-scan-and-dfdy-eta`;
retarget as the stack merges. It cannot be based on `main` — it rewrites the
loop body PR 1 introduces and its context includes the scan PR 2 removes.

Found while checking whether anything else in that loop was being redone per
row. The loop re-wrapped **its own output columns** on every row —
`as<IntegerVector>(lst[0])`, `as<NumericVector>(lst[1])`, and so on, 16 sites,
roughly 8–9 per row. These are not coercions: the vectors were allocated as
the right type a few lines above, so each call is a no-op cast that still pays
Rcpp's preserve/release bookkeeping. Nothing about the pointers can change
across rows.

| | before | after | |
|---|---|---|---|
| solve, 160k rows | 1.068 s | 0.284 s | **3.8x** |
| solve, 320k rows | 2.717 s | 0.966 s | **2.8x** |
| `etTrans()` alone, 320k rows | 1.232 s | 0.455 s | **2.7x** |

I had estimated 5–10% before measuring, and was wrong by two orders of
magnitude — the preserve/release traffic dominated. It also reframes #1325:
the `double` column was never the floor it looked like, because it paid this
same overhead all along. The issue was two stacked costs — the quadratic
coercion only integer/logical paid, on top of a constant factor every solve
paid.

## Verification

A speedup that large is the shape of "you stopped writing the data", so
correctness was checked before the number was believed. On a case covering
double, time-varying-with-`NA`, integer and logical covariates, a `keep`
column and 200 subjects, `rxSolve()` output, `etTrans()` output and the per-id
covariate table all compare equal to the previous build.

Full suite: **0 failures, 0 errors, 39 270 passing, 5 skipped** — the same
counts as the base branch.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -719,6 +719,14 @@
   double covariate at 20000 subjects.  The columns are now coerced once, as
   above.
 
+- Translating an event table no longer re-wraps its own output columns once
+  per row.  The row loop in `etTrans()` took each output column as a fresh
+  `Rcpp` vector on every row -- a no-op cast, since the columns were
+  allocated as the right type, but one that still paid Rcpp's
+  preserve/release bookkeeping at every one of roughly nine sites per row.
+  The pointers are taken once instead, which is 2.7x on `etTrans()` alone and
+  2.8-3.8x on a solve at 160000-320000 rows.
+
 - `rxDfdy()` (and `rxModelVars()$dfdy`) report an ETA derivative as
   `df(A)/dy(ETA[1])` instead of leaking the internal name
   `df(A)/dy(_ETA_1_)`.  Both `THETA[n]` and `ETA[n]` are translated back from
--- a/NEWS.md
+++ b/NEWS.md
@@ -727,6 +727,11 @@
   The pointers are taken once instead, which is 2.7x on `etTrans()` alone and
   2.8-3.8x on a solve at 160000-320000 rows.
 
+- `etTrans(allTimeVar = TRUE)` no longer errors when an `iCov` covariate is
+  supplied.  Under `allTimeVar` every covariate is emitted as a per-row
+  column, but the column for an `iCov` covariate was never allocated, so
+  taking it failed instead of returning the covariate.
+
 - `rxDfdy()` (and `rxModelVars()$dfdy`) report an ETA derivative as
   `df(A)/dy(ETA[1])` instead of leaking the internal name
   `df(A)/dy(_ETA_1_)`.  Both `THETA[n]` and `ETA[n]` are translated back from
--- a/src/etTran.cpp
+++ b/src/etTran.cpp
@@ -3094,13 +3094,40 @@ List etTrans(List inData, const RObject &obj, bool addCmt=false,
     covDp[j] = REAL(curCovD);
     covDn[j] = curCovD.size();
   }
+  // The output columns below are ours -- allocated above as IntegerVector /
+  // NumericVector -- so re-wrapping them per row only pays Rcpp's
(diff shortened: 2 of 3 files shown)


mattfidler:
## Independent review (Antigravity / gemini-3.1-pro-high)

Ran an outside reviewer over this diff. It returned **one** finding; everything
else it explicitly signed off on, and I re-checked each of those:

- **Pointer lifetimes are safe.** `lst` / `lst1` / `keepL` are stack-local
  `Rcpp::List`s (rooted), R's GC does not relocate, the vectors are freshly
  allocated so not ALTREP, and nothing between the hoist and the last use
  reassigns an element or reallocates.
- **The index arithmetic that replaced the running `kk` counter is exact.**
  `baseSize` is `6+cmtAdd` when `CENS` is allocated and `6+cmtAdd+censAdd` when
  `LIMIT` is, so `outCens` / `outLimit` land on the same elements `kk++`
  reached.
- **Guards match.** The hoist's `hasCmt && covCol[j] >= 0 && j == cmtI` and
  `!allTimeVar && covCol[j] < 0` are the same conditions the loop body uses, and
  `keepL[j]` / `lst1[1+j]` are always the type `REAL()` assumes.

### The one finding, confirmed

`src/etTran.cpp:3122`. With `allTimeVar = TRUE` and an `iCov` covariate
(`covCol[j] < 0`), the allocation block never assigned `lst[baseSize+j]`, so it
stayed `R_NilValue`:

```r
etTrans(d, m, allTimeVar = TRUE, iCov = data.frame(id = 1:3, wt = c(70, 80, 90)))
#> Error: REAL() can only be applied to a 'numeric', not a 'NULL'
```

One correction to the reviewer's account: it claimed the pre-diff code silently
wrote out of bounds into the R heap. It does not -- checked against installed
5.1.7, where `as<NumericVector>(R_NilValue)` throws `Not compatible with
requested type: [type=NULL; target=double]`. So this is a **pre-existing broken
path, not a regression and not memory corruption**; this PR only changes which
error is raised. `nlmixr2est` passes `allTimeVar=TRUE` in `saem_fit`, `focei`
and `augPred`, but never alongside `iCov`, so nothing downstream reaches it.

Fixed in 167190a53. The `covColj < 0` branch now allocates a full-length column
under `allTimeVar`, replacing a dead `lst1[1+j]` assignment that was overwritten
a few lines later:

```cpp
if (allTimeVar) {
  lst[baseSize+j] = NumericVector(idxOutput.size()-rmAmt);
} else {
  lst[baseSize+j] = NumericVector(0);
}
```

Since the path only ever errored, this cannot regress a working call. Added a
regression test in `tests/testthat/test-iCov.R`, plus `NEWS.md` entries for both
the fix and the pointer hoist itself, which had none.

### Verification

`etTrans|cov|keep|iCov|geom-cens|solve|event`: **0 failures, 3608 passing** (2
warnings, both ggplot2 deprecations unrelated to this change).

billdenney:
That review suggests that both the immediate finding should be fixed and also that iCov should get the same once-over conversion treatment on input. Right?

mattfidler:
Probably, though I'm unsure if that it would be affected, since I think that it may have already be transitioned to a double matrix.  But it wouldn't hurt.

I simply don't use `iCov` myself very often for it to have done much for me.

As a note, `iCov` in general is a bit faster than the regular covariate path, so the differences may be smaller.

PR #1335, opened 2026-09-02
Prevent transformation on variables during confint plot

OmarAshkar:
`plot.rxSolveConfint` allows for unquoted simulated variable name. Unexpected results happens if a user tries to apply transformation on that variable, which is intuitive. To avoid that issue, this PR enforces no transformation at this step all together.

### DIFF
--- a/tests/testthat/test-confint.R
+++ b/tests/testthat/test-confint.R
@@ -164,15 +164,11 @@ rxTest({
     .ci <- suppressMessages(confint(.s, "cp", level=0.95, ci=FALSE))
 
     .p <- plot(.ci, cp)
-    .p2 <- plot(.ci, cp/10)
-    .p3 <- plot(.ci, log(cp))
+      expect_true(inherits(.p, "ggplot"))
 
-    expect_true(inherits(.p, "ggplot"))
-    expect_true(inherits(.p2, "ggplot"))
-    expect_true(inherits(.p3, "ggplot"))
-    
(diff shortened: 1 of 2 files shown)


mattfidler:
Hi @OmarAshkar 

I'm really unsure what the use case of this is.  Also, all the tests fail.  

Matt

tests/testthat/test-confint.R
```diff
@@ -156,4 +156,22 @@ rxTest({
     expect_true(all(c("p1", "eff") %in% names(.ci)))
   })
 
+  
+  test_that("plot.rxSolveConfint1 works with output tranformation", {
+    rxSetSeed(42)
```
mattfidler: Use `rxWithSeed()` instead so that this doesn't leak the RNG state to other places.

mattfidler (commented):
Still don't understand the use case from the code.

OmarAshkar:
Hi @mattfidler 
The problem I had is setting transformation on `plot.rxSolveConfint1` was not either applied no failed.  This PR initially attempt to make a turnaround to apply the parsing if transformation is detected. But now, I feel it is better to add more strictness to avoid transformations at this step all together. 

```
 res <- rxSolve(mod, et, nSub=1000) |> confint(cp, level=0.95, ci=FALSE)
plot(res, cp/10) # This used to run without errors. 
```

mattfidler:
I suppose that makes sense

Issue #105, opened 2022-03-07
add `rename` ability


mattfidler:
This allows rxode ui models to rename parameters, using the `dplyr` s3 method `rename`

billdenney:
The work on this may be tangentially related to #72 and #73.

mattfidler:
Overall yes, though I am unclear about this:

> and in the data? 

My guess is you are talking about the compartment specifications needed in the data itself.  If a compartment is renamed then it corresponding data requirement is also renamed.  Compartment numbers are unlikely to change unless you change a residual to match a compartment.

billdenney:
For the data, I'm not clear that it's a good idea to modify the user's data-- it was mostly for consideration.

The idea of "and the data" would be that if we have a model fit for two PK models and want to then combine them into a multiple PK, single PD model, for instance, we may want to carry the data through the process and modify the compartments as we go.  Or, we may not want to do that because it could be error-prone (and it could possibly include data that the user doesn't intend to include).  Or, we may just want to stay away from automatic data modifications, regardless.

I do think that the user probably would need to use compartment naming rather than numbering for it to work well.

mattfidler:
No only models, no data modification.

PR #1344, opened 2026-09-02
perf(etTrans): test id membership with a set, not a linear scan- #1337

billdenney:
Stacked on #1336; retarget to `main` once that merges.

Found by sweeping for the pattern behind #1325 rather than waiting to trip over
it again. This is a **third** instance in the same function, and larger than the
pointer hoist.

`allId`, `obsId`, `zeroId` and `doseId` hold one entry per subject, and whether
an id was already present was tested with `std::find()` **once per input row** —
12 sites in the row loop, plus one more per row when dropping dosing-only ids.
That makes translation `O(rows × subjects)` instead of `O(rows)`.

Holding the row count **fixed at 120 000** and only redistributing it over more
subjects, so the only thing changing is the id count:

| subjects | before | after |
|---|---|---|
| 100 | 0.025 s | 0.023 s |
| 1 000 | 0.041 s | 0.016 s |
| 6 000 | 0.163 s | 0.019 s |
| 12 000 | **0.316 s** | **0.021 s** |

12.6x growth becomes flat; **15x at 12 000 subjects**. This matters
independently of everything already fixed, because it scales with the subject
count rather than the data size — a large population simulation pays it on top.

## Approach

The vectors stay: their `size()` is compared, and the "IDs without observations
dropped" warning iterates them in order. An `unordered_set` alongside each
carries the membership test, inserted into at every `push_back`. Also reserves
`zeroId`, which was never reserved — that line reserved `obsId` a second time.

## Verification

Output **and warning text** identical, checked across the paths those vectors
drive: ids with no observations, all-zero ids, dosing-only ids, `keepDosingOnly`
both ways, plus a 300-subject randomised case.

Regression test in `test-etTrans.R` verified to go **red on the base branch and
green here** — it holds rows fixed and varies only the subject count, so it
cannot pass by accident.

Replaces #1337 with the same code. But this time, with the target branch of main


mattfidler:
Merged main to fix the failing checks: the `test-ss-extra-dose-duration.R` failure was pre-existing on `main` (already fixed there via a different assertion, unrelated to this change) -- not caused by this PR. Full local suite after the merge: `[ FAIL 0 | WARN 38 | SKIP 5 | PASS 40604 ]`.

**Performance claim re-verified** after the merge, with optimized (`-O3`) `R CMD INSTALL` builds into separate scratch libraries (never `load_all`, which builds at `-O0`), interleaving pre-fix/post-fix runs to rule out machine noise:

| subjects | pre-fix (std::find) | post-fix (unordered_set) |
|---|---|---|
| 100 | 0.022s | 0.020s |
| 12000 | 0.239s | 0.021s |

Pre-fix scales ~11x from 100 to 12000 subjects (same row count, just spread over more ids); post-fix is flat, for an ~11x speedup at 12000 subjects -- consistent with the commit's claim of ~15x.

**Independent review via Antigravity (gemini-3.1-pro-high)**: no correctness, memory-safety, or concurrency issues found -- every `std::find()` on `allId`/`obsId`/`zeroId`/`doseId` is converted to a set lookup, and every `push_back` has a matching `insert` keeping vector and set in lockstep. It flagged that the pre-existing "IDs without zero-time start" warning path has no test asserting the warning string specifically; on inspection that code path (including the `zeroIdSet` lookup this diff touches) is already exercised by the existing "X(0) should be at time zero; see issue #105" test, which checks the resulting TIME/EVID values rather than the warning text, so no functional gap -- not adding a redundant test.

mattfidler:
Windows CI caught a real flaw in the new timing test itself, not in the fix:

```
Expected `.minElapsed(.many)` < `3 * .minElapsed(.few)`.
Actual comparison: 0.010 >= 0.000
```

`.minElapsed(.few)` measured 0.000s -- a single `etTrans()` on that arm ran faster than that runner's timer resolution, so `3 * 0 = 0` and the `.many` arm's real, non-zero, and perfectly fine time failed the comparison outright. The performance fix itself was never in question (see prior verification above); only the test's per-call timing was unreliable.

Fixed by summing a batch of calls (`reps = 20L`) inside one `system.time()` per sample, so the measured interval is always well above whatever clock granularity a runner has, on any platform. Confirmed locally (even under a slow `-O0` `load_all` debug build): few=0.154s, many=0.193s over the batch -- both comfortably measurable, ratio ~1.25x, well under the 3x threshold.

mattfidler:
Merged main again (it had moved 29 commits, including the "one event translator" refactor, PR #1355, which restructured exactly the code this fix touches). Non-trivial conflict in `src/etTran.cpp`: main replaced the inline record-push/zeroId-tracking code in the generic-dose branch with new steady-state-duration validation plus a call into the shared `_rxTranslateDoseInto()`, and that refactor also added a brand-new zeroId membership check (in the per-occurrence `addl` loop) that didn't exist when this PR's fix was written, still using the old `std::find()` linear scan. Resolved by keeping main's new validation/translator-call structure and converting that new site to `zeroIdSet` as well, so the fix now covers every membership test on `allId`/`obsId`/`zeroId`/`doseId`, including the one main added after this branch's original commit.

Verified after merge:
- No leftover `std::find()` on any of the four id vectors.
- `test-etTrans-golden.R` (~14300 cases, exact-output snapshot) and `test-etTrans-translator.R` (event-table vs. runtime-push agreement) both pass -- the two gates CLAUDE.md requires for any change here.
- Full local suite: `[ FAIL 0 | WARN 39 | SKIP 5 | PASS 59053 ]`.

**Performance re-verified against the new baseline** (origin/main tip, `80b83cf3e`, vs. this branch), same methodology as before (`-O3` scratch installs, interleaved runs):

| subjects | main (pre-fix) | this branch (post-fix) |
|---|---|---|
| 100 | ~0.020s | ~0.019s |
| 12000 | ~0.24s | ~0.021s |

Still an ~11-12x speedup at 12000 subjects -- the win survived the refactor merge, including for the new call site the merge picked up.

PR #1356, opened 2026-09-04
build: track header dependencies so an object cannot go stale

billdenney:
## What

Compile with `-MMD -MP` and read the generated `.d` files back in
`src/Makevars*`, so an object is rebuilt when a header it includes changes --
not only when its own source changes.

## Why

R's build rules make an object depend on its own source alone. `make`
therefore reuses `foo.o` whenever only a header that `foo.cpp` includes has
changed, and nothing warns.

That crosses package boundaries. A `LinkingTo` package's headers are just
another include path, so when a struct there gains or reorders a field, the
objects compiled against the previous layout are kept and linked alongside
freshly compiled ones. The library's halves then disagree about member
offsets -- a corrupt binary rather than a compile error, surfacing as a
segfault far from the cause.

This is not hypothetical. On a development install of rxode2 5.1.7:

- `src/approx.o` was compiled 2026-07-24.
- `inst/include/rxode2parseStruct.h`, which defines `rx_solving_options_ind`
  and `rx_solving_options`, changed 28 times after that, twice reshaping the
  structs outright ("append the origin fields at the end of their structs",
  "struct tail").
- `approx.cpp` itself had not changed since 2026-06-15, so `make` never
  recompiled it. 129 of 270 objects in that tree were older than the newest
  header; in nlmixr2est, 15 of 34 were three months old.

`_update_par_ptr()` and `getValue()` in `approx.cpp` then read
`ind->cov_ptr`, `ind->ix` and `op->ncov` at obsolete offsets. They run only
when a solve carries a per-row covariate, so most work was unaffected and the
failure looked like an estimator bug:

```r
library(rxode2)
m <- rxode2("param(WT); y = 2 * WT;")
d <- data.frame(ID = rep(1:2, each = 3), TIME = rep(0:2, 2),
                WT = c(70, 71, 72,  80, 80, 80))
rxSolve(m, d)   # segfault; one subject, or a WT constant within subject, is fine
```

Every nlmixr2est SAEM fit whose model read a covariate segfaulted --
`.configsaem()` hands the kernel a pre-translated event matrix, so the
covariate stays a per-row column -- while focei on the same model was fine,
because `etTrans()` reduces an id-invariant covariate into the per-id `cov1`
table and `ncov` is 0. nlmixr2est's own `tests/testthat/test-inpar-saem.R`
segfaulted as well. A clean rebuild fixed all of it: no source was wrong.

Documentation cannot prevent this, and a green test suite does not catch it,
because the suite runs against the same mixed binary.

## How

`-MMD` makes each object record its headers in a `.d` file beside it; `-MP`
emits a phony target per header so one that is later renamed or deleted does
not break the next build. The `.d` files the previous build wrote are read
back at the end of `src/Makevars*`.

`.DEFAULT_GOAL := all` is load-bearing, not cosmetic. `Makevars` is read
before `shlib.mk` defines `all`, so without it the first target of the first
included `.d` file becomes make's default goal: the objects rebuild and the
shared library is never relinked -- failing silently in exactly the way this
change exists to prevent. This was observed, not theorised.

`.d` files are added to `.gitignore`, to `.Rbuildignore`, and to the
`cleanup` scripts (which run for `R CMD build` and `R CMD INSTALL --clean`,
not on an ordinary install, so incremental builds keep their dependency
information).

## Verified

Per package: clean build, then a no-op rebuild, then touching one header.

| package | objects / depfiles | no-op rebuild | header touched | recompiled | relinked |
| --- | --- | --- | --- | --- | --- |
| rxode2 | 270 / 263 | 22 before the second commit below | `rxode2parseStruct.h` | 50 | yes |
| nlmixr2est | 34 / 34 | 0 | `armahead.h` | 16 | yes |
| babelmixr2 | 5 / 5 | 0 | `timeIndexer.h` | 1 | yes |
| nonmem2rx | 19 / 19 | 0 | `abbrec.g.d_parser.h` | 1 | yes |
| rxode2ll | 17 / 17 | 0 | `llik.h` | 16 | yes |
| nlmixr2extra | 3 / 3 | 0 | (no in-package header is included) | -- | -- |

rxode2's 263 depfiles for 270 objects are correct: the 7 Fortran sources are
not given `-MMD` (they have no `#include` graph to track).

## Second commit: stop regenerating unchanged sources

rxode2 only. `inst/tools/workaround.R` rewrote every file it generates on
each install -- the vendored SUNDIALS sources, `RcppExports.cpp`, `sbuf.c`,
`codegen2.h`, `dfIni.R`. Rewriting a file byte-for-byte still gives it a new
mtime, which is why a no-op rebuild recompiled 22 objects before this branch,
and why tracking header dependencies would otherwise have spread that churn
to everything including the regenerated headers.

`inst/include/rxode2parseVer.h` was worse than churn: it holds an md5 digested
from every file in `src/` and `inst/include` -- itself included -- so each
build folded the previous build's fingerprint into the new one:

```
run 1: __VER_md5__ "cd3502928e7fa0b78a8e49f954aff1e9"
run 2: __VER_md5__ "a931d9be31b9c3605b631fcfd3b5bc29"   # nothing else changed
```

That header is included by `genModelVars.h`, `rxData.cpp` and `rxode2_df.cpp`,
so the churn reached well past the file itself -- and the md5 could never key
the model cache it exists for, since a fingerprint that changes every build
never matches. It is now excluded from its own digest.

`.writeLinesIfChanged()` compares before writing; every one of these writes
already used a binary connection, so the comparison is exact. The four
`unlink()` calls that deleted a generated file just before rewriting it are
gone -- they would have defeated the comparison.

Verified: three consecutive `workaround.R` runs change no mtimes at all.

A throwaway two-source package confirms the mechanism end to end: with a
header changed from `111` to `222` and no source touched, the old build
recompiles nothing and keeps returning `111`; with this change it recompiles
only the one dependent source, relinks, and returns `222`.

## Portability

`-MMD -MP` are GCC/Clang spellings, understood by every compiler CRAN
currently checks with (GCC, Clang, Intel oneAPI's clang-based `icx`). If you
would rather not take even that risk, the flags can be gated behind a
compiler probe in `configure`/`workaround.R`; say the word and I will add it.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -166,3 +166,4 @@ tests/testthat/test-lincmt-parse.R
 ^LICENSE\.md$
 ^tools$
 ^tools/
+^src/.*[.]d$
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,4 @@ tests/testthat/_problems/
 *.tar.gz
 *.tgz
 inst/tools/*.Rout
+/src/*.d
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,16 @@
 # rxode2 5.1.7 (development version)
 
+## Internal
+
+- The build now tracks header dependencies.  R's rules rebuild an object only
+  when its own source is newer, so an incremental build silently kept objects
+  that had been compiled against an earlier version of a header they include.
+  That reaches across packages: a `LinkingTo` package's headers are just
+  another include path, so a struct whose layout changes there leaves objects
+  here compiled to the old layout and the resulting shared library mixes both
+  -- a corrupt binary rather than a compile error.  `src/Makevars*` now
+  compiles with `-MMD -MP` and reads back the generated `.d` files.
+
 ## Bug fixes
 
 - A steady-state constant infusion (`ss=1`, `ii=0`, `amt=0`) pushed from
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,15 @@
   -- a corrupt binary rather than a compile error.  `src/Makevars*` now
   compiles with `-MMD -MP` and reads back the generated `.d` files.
 
+- `inst/tools/workaround.R` now writes a generated source only when its
+  content actually changed.  It had rewritten the vendored SUNDIALS sources,
+  `RcppExports.cpp`, `sbuf.c`, `codegen2.h` and the rest on every install,
+  refreshing their mtimes and so recompiling 22 files each time.  Relatedly,
+  `inst/include/rxode2parseVer.h` was digested into the very md5 it carries,
+  so that fingerprint -- and the header -- changed on every build regardless
+  of the sources; it is now excluded from its own digest, which also lets the
+  model cache the md5 keys actually hit.
+
 ## Bug fixes
 
 - A steady-state constant infusion (`ss=1`, `ii=0`, `amt=0`) pushed from
--- a/cleanup
+++ b/cleanup
@@ -3,3 +3,4 @@ rm -f src/Makevars
 rm -f config.*
 rm -f src/Makevars.win
 rm -f inst/include/*.gch
+rm -f src/*.d
--- a/cleanup.win
+++ b/cleanup.win
@@ -3,3 +3,4 @@
 rm -f src/Makevars.win
 rm -f src/Makevars
 rm -f inst/include/*.gch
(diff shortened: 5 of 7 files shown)


mattfidler:
Again, with strict file name requirements, the `.d` likely won't pass cran.

PR #1357, opened 2026-09-04
abi: walk the subject array at its allocated stride, not rx2api.c's sizeof

mattfidler:
Walks the subject array at the stride it was allocated with.

`src/rx2api.c` is the package's ABI surface -- the file downstream packages
resolve through the external-pointer table -- and it is a *different*
translation unit from the one that allocates `rx->subjects`
(`rxOptionsIniEnsure`, `src/par_solve.cpp`). `getSolvingOptionsInd()` returned
`&rx->subjects[id]`, which compiles that TU's own
`sizeof(rx_solving_options_ind)` into the pointer arithmetic, so a correct
subject pointer needed the two TUs to agree on the layout.

They do not have to. R's default make rules track no header dependencies, so an
object file whose own source did not change is linked in unchanged after a
change to the struct. That is what happened on a local build: `rxData.cpp` and
`par_solve.cpp` saw 1664 bytes and `rx2api.c` saw 976, and every subject but
index 0 was read 688*i bytes short of where it lives. Subject 0 sits at the
start of the array either way, which is why a single-subject dataset looked
fine and every multi-subject one did not.

Downstream, in nlmixr2est, that surfaced as `focei: dataset too large`, an
`R_Calloc` failure for an absurd size, or a segfault, depending on what the
mis-read bytes happened to hold (nlmixr2/nlmixr2est#1039).

## The change

`rxOptionsIniEnsure()` publishes the stride it allocates with, and
`getSolvingOptionsInd()` uses it. A plain global rather than a field of
`rx_solve`, because reading a field would put a layout back in the loop.
Fields keep their offsets by the append-only convention the struct already
documents, so the stride was the whole of the disagreement.

## Verification

Rebuilding only `rx2api.o` against a struct grown by 688 bytes, and running the
nlmixr2est `theo_sd` FOCEi reproduction from nlmixr2/nlmixr2est#1039:

| rxode2 | `rx2api.o` | result |
| --- | --- | --- |
| main | current | objf 133.5727 |
| main | stale | fails |
| this branch | current | objf 133.5727 |
| this branch | stale | objf 133.5727 |

## Tests

`tests/testthat/test-abi-subject-stride.R`, two of them:

- `rxTestAbiSubjectCounts_()` reads each subject's record count through the
  accessors downstream resolves and compares against the dataset, over subjects
  with unequal record counts (subject 0 reads correctly under any stride, so
  the test needs more than one subject to bite). It also reports the published
  stride next to the calling TU's own. It deliberately does NOT also read
  `rx->subjects[i]` directly: under exactly the staleness this exists to catch,
  the direct read is the one that runs off the end of the allocation, and a
  test should fail rather than read unallocated memory.
- `rxTestAbiStrideProbe_()` builds the disagreement on purpose, with no build
  trickery: a private subject array whose entries are `pad` bytes further apart
  than this build's `sizeof`, each marked, with the published stride set to
  match. Pre-fix it reads `1000 0 0 0 0 0` for `pad = 8` and `pad = 688`
  against markers `1000..1005`; `pad = 0` passes either way, which is the
  point.

## What this does not fix

Every other translation unit here -- the solvers, `rxData.cpp`,
`rxSerialize.cpp`, `rxode2_df.cpp` -- still subscripts `rx->subjects` at its
own compiled offset, and so does the C this package emits for a model. That is
the same hazard, but it is a hazard *within one build*: a stale object file
inside rxode2 is a broken rxode2, and no API contract defends against it. The
ABI surface is different because it is the contract another package resolves,
and it is the file most likely to go stale precisely because it changes rarely.

The generated model code cannot go stale across an upgrade: `rxMd5()` folds the
md5 of rxode2's own DLL into the model cache key, so a rebuilt rxode2
recompiles every cached model.

If the broader hazard is worth closing, the root fix is header dependency
tracking in `src/Makevars` (`-MMD -MP` plus `-include $(OBJECTS:.o=.d)`), which
would have prevented this outright. Not done here.

## Review

Four rounds of independent review (Antigravity, Gemini). Fixed from it: an
unreachable `rxIndSize == 0` fallback, the canary's out-of-bounds direct read,
and the absence of any test of which stride the accessor actually follows. The
final round reports no correctness, memory-safety or concurrency findings.

Companion: nlmixr2/nlmixr2est#1041 keeps a mismatched build from reaching the
allocator quietly, for anyone on a released rxode2 that predates this.


mattfidler:
@billdenney This may be a different solution to #1356

PR #1358, opened 2026-09-09
fix(mix): an expanded mix() is still a mixture model

mattfidler:
## The bug

A `mix()` model that has been through symengine -- which is **every** estimation
method's prediction model -- was no longer recognized as a mixture model at all.

symengine expands the `mix()` call into a sum of per-component selectors, so the
call itself is gone by the time the parser sees the model.  `tb.hasMix` is only
set by a literal `mix()` call, so `RxMvFlag_mix` came back 0, and from there:

- `etTrans()` discarded a `mixest` column supplied in the data or in `iCov`;
- nothing populated `ind->mixest`, because that only happened inside `_mix()`,
  which an expanded model never calls;
- every `mixest == k` selector was therefore false, and each `mix()`-derived
  variable solved as **0**.

In a fit table that is silent: the predictions are wrong, not missing.  Reported
downstream as nlmixr2/nlmixr2est#1041, where a SAEM mixture fit came back with
`mixest`, `mixnum` and the whole `mix()` result zero for every subject, and
`IPRED`/`PRED` computed from an elimination rate of 0.

## The fix

The expansion now emits one reserved `rx_mixsel_<k>_<n>_` selector per
component -- "the k-th of n" -- and the parser turns that back into the
`mixest == k` comparison while recording the component count.

The total is **spelled out** rather than inferred from the largest selector
present, because a component whose expression folds to zero (any sensitivity
with respect to an eta only one component uses) drops its selector out of the
expression entirely.  Selectors that disagree with each other, or with a literal
`mix()` in the same model, about the count are a syntax error.

The selector goes into symengine as an ordinary symbol.  It is constant in every
eta, so it rides through differentiation untouched and comes back out unchanged
-- which means neither `rxFromSE()` nor its C fast path needs a rule for it, and
they cannot disagree.

Three further gaps in the same path are closed:

- `ind->mixest` is decoded from the supplied value at individual setup rather
  than only inside `_mix()`.
- A `mixest`/`mixunif` column in `iCov` now splits a homogeneous solve group.
  Subjects sharing an event table solve as one group, and the group was only
  split on iCov columns that are model parameters; `mixest` is a reserved
  variable, so the whole group took the first subject's component.  Only a
  mixture model splits on it, so a non-mixture model's grouping is unchanged.
- The grouped path indexes the per-id `mixest` vector by group rather than by
  the expanded subject counter, which previously walked off the end of it.

## An unrelated bug found on the way

`.etGroupedSolveSplitKey()` fed `interaction()` the raw columns.  `interaction()`
is `NA` for a row with an `NA` in any column and the `split(drop = TRUE)` it
feeds **discards** that row -- so a subject with a missing value in a column its
group is split on vanished from the solve output entirely, rather than being
rejected.  `NA` now gets a level of its own and the subject reaches the
per-subject validation that should reject it: a missing `mixest` is now the
error it always should have been.

## Behaviour changes

- `rx_mixsel_<k>_<n>_` is a new reserved variable name, alongside `mixest`,
  `mixnum` and `mixunif`.
- `rxToSE()` on a `mix()` call now emits the selector instead of
  `rxEq(mixest, k)`.
- The expansion drops the mixture *probabilities* along with the `mix()` call,
  so an expanded model can be told which component a subject belongs to
  (`mixest`) but cannot sample one from a supplied `mixunif`.  Simulation from
  `mixunif` needs the `mix()` call, which every hand-written model keeps.

## Testing

`test-mix.R` gains coverage for: the expanded form reporting as a mixture and
its selectors not becoming parameters; the normalized-model round trip; names
that only look like selectors staying ordinary variables; a folded component not
shrinking the mixture; disagreeing counts erroring; `mixest` per individual
through both `iCov` (on a homogeneous event table, which is what exercises the
group split and the group indexing) and a plain data column; and `NA`/`0`/`> nMix`
being rejected rather than silently dropped or zeroed.

`test-mix.R`, `test-iCov.R`, `test-et.R` and the symengine translation tests
(including the byte-for-byte `rxFromSE` fixture) pass locally; the full suite is
left to CI.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -31,10 +31,13 @@
 - A `mix()` model whose call has been expanded by symengine -- the form
   every estimation method's prediction model is built from -- is now still
   recognized as a mixture model.  The expansion emits one reserved
-  `rx_mixsel_<k>_` selector per component, which keeps the component count
-  the dropped `mix()` call carried, so `mixnum` reports it and a
+  `rx_mixsel_<k>_<n>_` selector per component, spelling out the component
+  count the dropped `mix()` call carried, so `mixnum` reports it and a
   per-individual `mixest` supplied in the data or in `iCov` reaches the
-  solve.  Previously such a model parsed with no mixture at all: the
+  solve.  The total is spelled out rather than inferred from the largest
+  selector present, because a component whose expression folds to zero --
+  any sensitivity with respect to an eta only one component uses -- drops
+  its selector out of the expression entirely.  Previously such a model parsed with no mixture at all: the
   `mixest` column was discarded, `ind->mixest` stayed 0, and every
   `mix()`-derived variable solved as 0 -- which silently corrupted the
   predictions in the fit table (nlmixr2/nlmixr2est#1041).
--- a/NEWS.md
+++ b/NEWS.md
@@ -55,6 +55,12 @@
 - `rx_mixsel_<k>_<n>_` is now a reserved variable name, like `mixest`,
   `mixnum` and `mixunif`.  A model cannot use it for anything else.
 
+- An `iCov` column that a homogeneous solve group is split on no longer
+  drops the subject when its value is `NA`.  The split key came from
+  `interaction()`, which is `NA` for a row with any `NA`, and the `split()`
+  it feeds discarded that row -- so the subject vanished from the solve
+  output instead of being rejected.
+
 ### Event translation
 
 - A steady-state dose into a compartment with a modeled `alag()` pushed
--- a/NEWS.md
+++ b/NEWS.md
@@ -53,7 +53,15 @@
   whole group took the first subject's component.
 
 - `rx_mixsel_<k>_<n>_` is now a reserved variable name, like `mixest`,
-  `mixnum` and `mixunif`.  A model cannot use it for anything else.
+  `mixnum` and `mixunif`.  A model cannot use it for anything else, and
+  selectors that disagree with each other, or with a literal `mix()` in the
+  same model, about the number of components are a syntax error.
+
+- Note that the expansion drops the mixture PROBABILITIES along with the
+  `mix()` call, so an expanded model can be told which component a subject
+  belongs to (`mixest`) but cannot sample one from a supplied `mixunif`.
+  Simulation from `mixunif` needs the `mix()` call itself, which every
+  hand-written model keeps.
 
 - An `iCov` column that a homogeneous solve group is split on no longer
   drops the subject when its value is `NA`.  The split key came from
--- a/NEWS.md
+++ b/NEWS.md
@@ -47,10 +47,13 @@
   never ran `_mix()`, so the supplied assignment never reached it.
 
 - A `mixest` or `mixunif` column in `iCov` now splits a homogeneous event
-  group.  Subjects that share an event table are solved as one group, and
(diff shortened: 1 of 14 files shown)


mattfidler:
## Antigravity review summary

Nine rounds against `agy` (`gemini-3.1-pro-high`), each a single narrow question
rather than one broad "review this". **Five findings were real and are fixed
in this branch**; the rest were verified against the code and dismissed.

### Real findings, all fixed here

1. **The component count was inferred as `max(k)` over the selectors present.**
   That undercounts whenever the last component's expression folds to zero --
   which is every sensitivity with respect to an eta that only one component
   uses. Fixed by spelling the total out in the name (`rx_mixsel_<k>_<n>_`)
   instead of inferring it. This is the finding that reshaped the design, and
   it also let the `rxFromSE` special case and the C emitter decline go away.

2. **Splitting a solve group on `mixest`/`mixunif` unconditionally** fragmented
   the event groups of any *non-mixture* model whose `iCov` happened to carry
   such a column, for a column that model never reads. The splitter went back
   to "columns the model uses"; the caller says the mixture names count as used
   only when the model is a mixture.

3. **An `NA` in an `iCov` column a homogeneous group is split on silently
   dropped the subject from the solve.** `interaction()` is `NA` for a row with
   any `NA` and the `split(drop = TRUE)` it feeds discards that row. This one is
   a pre-existing bug unrelated to mixtures that this branch would have started
   exposing; `NA` now gets a level of its own and the subject reaches the
   validation that should reject it.

4. **A literal `mix()` and the selector form could disagree about the count
   silently** -- `genModelVars()` reports `mix()` first. Both directions are now
   a syntax error.

5. `exists()` searches enclosing environments by default, so a same-named
   binding up the chain would suppress the local symengine registration
   (`inherits = FALSE`).

### Verified and dismissed

Each of these was checked by reading the actual code path or by running the
case, not by judging plausibility:

- *"`_mix()` may rely on `ind->mixest == 0` for first-pass init"* -- `_mix()`
  branches on `ind->mixunif`, and never reads `ind->mixest` before assigning
  it. Pre-setting it changes nothing.
- *"The other two `rxSetIndMix()` call sites still use `nsub`, so a grouped
  solve overwrites the value / runs off the end"* -- raised twice. Those two
  sites are in the non-grouped `else` branch, where one translated subject *is*
  one real subject; `nsub` is correct there and is incremented after the call,
  not before.
- *"`mixest = 0` requests sampling and now evaluates to 0"*, *"`mixest = NA`
  is silently accepted"*, *"`mixest > nMix`"* -- `etTrans` rejects all three.
  Confirmed by running each: `mixest must be between 1 and 2 for this model`,
  `mixest must be an integer value`. (The `NA`-via-`iCov` case was the one
  exception and became finding 3 above.)
- *"`mixest` supplied as a data column instead of `iCov` avoids the group
  split"* -- ran it; the per-subject component is correct.

### Rounds that came back clean

Memory safety and parser state across successive compiles; split-key semantics
across every column type a data frame can hold (grouping is unchanged for
factors, ordered factors, logicals, integers, doubles and dates -- only `NA`
handling changes, which is finding 3); and the public-contract round, which
surfaced no bugs and only the intended behaviour changes already listed in
`NEWS.md`.

mattfidler:
This is an odd way of specifying things, but I guess it works and satisfies everything needed.

PR #1366, opened 2026-09-12
Let a model be solved compacted to its leading states

hiddevandebeek:
Companion to nlmixr2/nlmixr2est#1082 (the FOCEi combined inner+outer build).

## What this does

A downstream package can already ask for a solve narrower than the model (`ind->neqOverride`, `rxEffNeq()`), but the generated code always wrote every derivative and read every state, so compacting a model against its own wider code overran the integrator's `ydot` (sized by the compacted neq) and read past the compacted state row. This makes it safe:

- `parseDdt.h`: each `d/dt()` line is emitted as `if (__DDTk__ < _rxNeqEff_) __DDtStateVar__[k] = ...`, so a compacted solve neither writes nor evaluates the derivatives beyond its width (for a sensitivity model those trailing lines are the bulk of the RHS).
- `codegen.c`: every generated function reads a state beyond `rxEffNeq()` as 0 (`_rxNeqEff_ = rxEffNeq(_ind, _solveData->op)` before the state loads), and the `state(0)=` initial-condition writes are guarded the same way.
- `rxDydtCompact()` reports the capability so nlmixr2est can pick the build at model-generation time.

Generated code for an uncompacted solve is unchanged in behaviour (the guards are always true there). The `dydt` prologue/epilogue and the `neq` array handed to the integrator are untouched -- a first version assembled the derivatives in a full-width local and copied `_neq[0]` of them out, which broke mixed ODE + `linCmt()` models because rxode2 hands `dydt` an `_neq[0]` that counts the linCmt states while the liblsoda buffer does not; that is why the per-line guard is the right shape.

## Why

nlmixr2est's `fast=TRUE` fit now carries the outer-gradient sensitivities on the inner model (one solve pool model, nothing swapped during a fit). Its inner iterations integrate only the first-order block of that model and one full-width solve per subject serves the outer gradient and Hessian; this codegen change is what makes that compacted inner solve legal.

## Tests

`test-event-sensitivities`, `test-event-sensitivities-api`, `test-logical`, `test-sens3`, `test-dde-sens`, `test-adjoint-sens`, `test-lincmt-solve-sens`, `test-cvode` pass; nlmixr2est's fast-gradient / outer-Hessian / jump-sensitivity suites pass against this build with objectives identical to the uncompacted solve.

https://claude.ai/code/session_01HmazN3HXkZHX92zvqLZUTm


hiddevandebeek:
@mattfidler like i said, not a 100% sure about it, but if this in the direction you want ill do some polishing to make it a bit nicer

hiddevandebeek:
Closing -- only needed by nlmixr2/nlmixr2est#1082, which is being redone.

PR #1367, opened 2026-09-13
mu: name the clashing etas when one expression carries two of them

billdenney:
Investigating a report that a between-subject eta, an inter-occasion eta and a covariate effect in one exponential silently lost its mu-referencing:

```r
vc <- exp(lvc + etaVc + etaVcOcc + eSiteVc * SITEFLAG)
```

**It does not.** On `origin/main` that line mu-references: `etaVc` is paired with `lvc`, the covariate is recorded in `$muRefCovariateDataFrame`, `etaVcOcc` keeps `curEval="exp"`, and `$nonMuEtas` is empty. It matches the two-line `vcBase` split on every object, and holds across the term orders, `SITEFLAG * eSiteVc`, a second covariate, and a parameter whose occasion-level effect carries no covariate. Nothing tested any of that, so this PR adds the tests.

Worth recording, because it makes "the eta is missing from `$muRefDataFrame`" a misleading signal: an occasion-level eta is **never** a `$muRefDataFrame` row. Rows are written at exactly one place in `R/mu.R` with `level="id"` hard-coded; IOV etas live in `$level`. Lost mu-referencing is `$nonMuEtas` being non-empty, which is also all `assertRxUiMuRefOnly()` checks.

### The message the report actually hit

The reported output is reproduced exactly by declaring the occasion eta without its level (`etaVcOcc ~ 0.1` rather than `etaVcOcc ~ 0.1 | OCC`). Both etas are then subject-level, one expression cannot mu-reference both, and the error said only:

```
currently do not theta + eta1 + eta2
```

which names nothing — so the only advice reaching the user was the generic "try putting the mu-referenced expression on a simple line". It now names the population parameter and both random effects in the order they were written, and gives both fixes:

```
cannot mu-reference 'lvc' against 2 subject-level random effects in one expression: 'etaVc', 'etaVcOcc'
mu-referencing pairs a population parameter with one random effect per level of variability
- if 'etaVcOcc' is inter-occasion variability, declare its level in `ini({})`, ie `etaVcOcc ~ 0.1 | OCC`
- otherwise keep one random effect on the mu-referenced line and combine the other in on a
  following line; for a log-normal 'vc' that is `vcBase <- exp(lvc + etaVc)` then
  `vc <- vcBase * exp(etaVcOcc)`
```

The giveaway that this is what happened, rather than an IOV parsing failure: the old warning named `etaVc` too. `lotriBaseCondition("OCC")` returns `"OCC"` (it only strips `:same:`), and `.muRefDowngradeCleanup()` filters to base condition `"id"` — so that warning can never name a genuine IOV eta.

### Dead guard in `.muRefSetNonMuEta()`

The guard keeping occasion-level etas out of `$nonMuEtas` tested `env$levels`, which is assigned **nowhere** in the package — the level names live in `env$info$level`. It never fired. No current caller reaches it with an occasion-level eta, so this changes no behaviour today; it stops the guard failing open later, and a test holds it to that (it goes red against the old guard).

I swept the `mu.R` / `mu-cov-downgrade.R` / `err.R` parse path for other `env$<field>` reads never assigned anywhere in `R/`. `env$levels` was the only one, and the check does flag it pre-fix.

### Testing

- Full suite: **58463 passing, 0 failed, 0 errors**, 5 skipped, 403 contexts.
- `test-mu.R`: 148 → 222 assertions.
- The 105 suite warnings are all pre-existing. Two mention the non-mu warning (`test-piping-ini.R`, `e9`–`e12`); running that file with the old guard patched back in gives identical results (2 warnings, 180 pass), so they are unrelated to this change.
- `R/mu.R` adds no new lints.

No model's meaning changes, and no exported API changes — `.muRefMultiEtaMsg()` is internal and `@noRd`.


mattfidler:
This is simply a more expressive error, no behavior change.

Issue #1218, opened 2026-08-10
Modelled-duration (RATE = -2) doses tied at the same TIME break start/stop pairing (data errors 686 + 886), even across different compartments


mattfidler:
Two `RATE = -2` (modelled duration) doses that fall at **exactly** the same `TIME` for the same subject break the start/stop pairing and the solve fails with data errors 686 + 886. This happens even when the doses go to **different compartments**, which is a perfectly legal dataset.

## Reprex

```r
library(rxode2)

m <- rxode2({
  ka <- 1; cl <- 3; v <- 30; D1 <- 2
  dur(depot)   <- D1
  dur(central) <- D1
  d/dt(depot)   <- -ka * depot
  d/dt(central) <-  ka * depot - (cl / v) * central
  cp <- central / v
})

ev <- rbind(
  data.frame(ID = 1L, TIME = 0, EVID = 1L, AMT = 150, RATE = -2, CMT = 1L, DV = NA_real_),
  data.frame(ID = 1L, TIME = 0, EVID = 1L, AMT = 150, RATE = -2, CMT = 2L, DV = NA_real_),
  data.frame(ID = 1L, TIME = c(1, 4, 8, 12, 24), EVID = 0L, AMT = NA_real_,
             RATE = NA_real_, CMT = 2L, DV = 0))

rxSolve(m, ev)
#> Recovered solving errors for internal ID 1 (576):
#>   Data error 686
#>   Data error 886
#> Error: could not solve the system
```

Staggering the two doses by `1e-8` solves fine, so it is purely the exact tie:

```r
ev2 <- ev; ev2$TIME[2] <- 1e-8
head(rxSolve(m, ev2), 3)
#>   time       cp      depot   central
#> 1    1 3.266579 47.4090334  97.99738
#> 2    4 7.507732  8.7765053 225.23195
#> 3    8 5.244521  0.1607468 157.33562
```

## Diagnosis

`576 = 64 + 512` = `rxErrModelData686 | rxErrModelDataErr886`.

Both bits are set by the adjacency checks in `inst/include/rxode2parseGetTime.h`:

- `handleTurnOnModeledDuration()` (`:201`) requires `isEvidModeledDurationStop(getEvidP1(ind, idx))` — the record immediately *after* a duration start must be its stop.
- `handleTurnOffModeledDuration()` (`:178`) requires `isEvidModeledDurationStart(getEvidM1(ind, idx))` — the record immediately *before* a stop must be its start.

[...]

PR #1374, opened 2026-09-14
Add splitInfusion(), splitInfusionBolus() and splitBolusInfusion() model directives (etTrans-time dose splitting)

mattfidler:
## Motivation

Monolix's double-absorption library lets ONE dose record feed two absorption routes with an `F1`/`1-F1` apportionment, and the routes can mix zero-order (`Tk0`) with first-order (`ka`) absorption. `splitBolus()` covers bolus-only splitting; building Monolix-parity double-absorption piping in nlmixr2lib (and exposing it in the nlmixr2shiny Model Builder) needs infusion-capable splitting.

## What's added

Two model directives, mirroring `splitBolus()`:

| directive | splits | notes |
|---|---|---|
| `splitBolus(cmt, ...)` (existing) | bolus records | also honored by `evid_()` runtime pushes |
| `splitInfusion(cmt, ...)` (new) | infusion records | data `RATE`/`DUR` (flags 1/2) and modeled `rate()`/`dur()` (flags 8/9) including their stop records (6/7, negative-amt data stops), type-preserving |
| `split(cmt, ...)` (new) | bolus AND infusion records | additionally PROMOTES a plain bolus record aimed at a target with a modeled `dur()`/`rate()` property to a modeled start/stop pair for that target — one bolus dose record can feed both a zero-order (infusion) and a first-order (bolus) path |

- **etTrans-time only**: unlike `splitBolus()`, the new directives do not rewrite `evid_()` doses pushed while solving (per design scope).
- At most one splitting directive per model (parser-enforced); target compartments must be unique.
- Promotion applies to regular boluses (`wh0==1`); steady-state boluses are copied as-is with a one-time warning when a target declares `dur()`/`rate()`.
- Promotion forces `hasModeledRateDur`, so the existing start/stop re-pairing pass (issue #1218 machinery) orders the emitted pairs.
- No R export named `split` (would mask `base::split()`); the directive is parsed from model text and documented in `?split-directive`. `splitInfusion()` gets an exported stub like `splitBolus()`.

## ABI/cache notes

- `rxModelVars` gains `splitInfusion`/`split` APPENDED at indices 33/34; genModelVars.c now emits named NULL placeholders at 31/32 that the R-side `$timeId<-`/`$md5<-` replace by name, so `RxMv_timeId`/`RxMv_md5` and all existing indices are unchanged (verified: hand-appended defines identical to regenerated `genDefine()` output).
- etTran reads of the new mv slots are size/type-guarded for cached mv objects from older builds.
- `parsed_md5` is unaffected for models not using the new directives (it digests model/ini/state/params/lhs text).

## Validation

- New tests in test-etTrans.R (8 blocks): data-RATE infusion split equivalence vs explicit multi-cmt event tables; bolus records untouched by `splitInfusion`; modeled `RATE=-2` split (start/stop pairs for both targets); `split()` bolus and infusion sources; promotion shape (start-before-stop pair to the dur target, plain bolus to the other, nothing left on the source); **solve-level equivalence**: a single bolus record through `split()` with `dur(central)`/`f` 0.7-0.3 solves identically (1e-6) to explicit modeled-duration central + bolus depot2 dosing; parser misuse errors; mv/normModel round-trip.
- Targeted suites green: etTrans/modelVars/evid-push filters — FAIL 0 | PASS 19309, then FAIL 0 | PASS 18773 on the etTrans filter with the new tests.
- Full suite running locally; CI will gate.
- Independent Gemini review (antigravity CLI) over the C and R diffs: no correctness findings (ndose accounting, pairing-pass ordering, memory pairing, and mv guards all verified).

## Grammar

`inst/tran.g` gains `splitInfusion_statement`/`split_statement` productions and the two identifier tokens; `src/tran.g.d_parser.h` regenerated via `dparser::mkdparse` (the documented workflow).


mattfidler:
Follow-up on the review round: the rename is in place and one real finding from re-review is now fixed.

**Renamed per discussion**: the generic `split()` is now `splitInfusionBolus()`, with a new mirror directive `splitBolusInfusion()`. Both split bolus *and* infusion records etTrans-time only.

**Order-dependent promotion (fixed in ceaae1de7)**: re-review correctly observed the two directives were byte-identical aliases — both promoted *every* `dur()`/`rate()` target, ignoring target order. Now `splitInfusionBolus()` promotes the FIRST target to the modeled infusion pair and `splitBolusInfusion()` promotes the LAST; every other target keeps its bolus copy even when it declares `dur()`/`rate()`. Docs, NEWS, and the SS-bolus warning updated to match, plus a distinguishing test (both targets declare `dur()`; asserts the two directives produce different translations and each promotes only its designated target).

PR #1081, opened 2026-06-15
General development branch merging many open PRs to simplify testing

billdenney:

PR #1380, opened 2026-09-15
feat(rxSolve): single=TRUE solves the event table in place

billdenney:
@mattfidler, I'm not sure if this is useful, but I'm nearly sure that it was lost in the noise of the rest that was going on in #1081.

Adds `rxSolve(single=TRUE)`, which solves the caller's event table in place instead of copying it into `_globals`.

Ported from the `dev-2026-06` integration branch ([#1081](https://github.com/nlmixr2/rxode2/pull/1081)), which has been open since June and is the only substantive thing still held there. Everything else on that branch has either landed independently, was reverted on the branch itself, or is superseded — see the bottom.

### What it does

The columns a solve only ever reads — time, amt, evid, dv, ii, cens, limit, and the covariates — are borrowed from the caller's data frame rather than copied. That is what the ALTREP col-view and CMT-translation classes are for: `etTransSingle()` is a zero-copy variant of `etTrans()` that hands back views onto the original columns, with a translated CMT column presented through an ALTREP wrapper.

Only the buffers the solve genuinely owns are allocated: a translated `idose`, and stand-ins for whichever of dv/ii/limit the data does not carry. Everything else points into R's own memory, so `gFree()` grows a `singleMode` arm that nulls the borrowed pointers instead of freeing them, and releases the retained data frame.

It builds directly on `src/rxode2_altrep.c`, which is already in main (`ce3c7216c`, "perf: ALTREP compact sequences for sim.id/id output columns") — this extends that file with the col-view and CMT-trans classes rather than introducing a competing implementation.

Defaults to `FALSE`, so nothing changes unless asked for. `single=TRUE` is rejected with `serializeFile`, which needs a snapshot the borrowed columns cannot give it.

### Two deliberate changes from the original

**`Rxc_single` is appended (142), not inserted.** The original put `single` after `dense` in `rxControl()`'s returned list. Those constants are generated positionally from `names(rxControl())`, so that shifts `Rxc_*` 115 through 141 — 27 constants — out from under every downstream package compiled against the old indices. Main's own source already says so:

```r
## appended at the end: the C++ side reads `rxControl` positionally
## through the generated `Rxc_*` defines, so a new field may only be
## added here, never inserted
```

so `single` goes there. The diff on both generated headers is a single added `#define`; no existing index moves.

**`gall_times_n` is carried into both arms.** Main gained allocation-count tracking for `gall_times` after this work forked, and the only two conflicting hunks were exactly that. The single-mode arm records the borrowed length; the copy arm keeps main's line unchanged.

### Testing

- `test-single-etTrans.R` (comes with the feature): **8/8**.
- `test-ui-drop-lines.R`: **41/41**.
- End to end on a 131-subject, 4239-observation dataset with a 2-compartment oral model: `single=TRUE` and `single=FALSE` produce output for which `identical()` is `TRUE`, and `single=TRUE` + `serializeFile` errors as intended.

Worth a full CI run — this touches `rxSolve_datSetupHmax` and `gFree`, and the memory-ownership split is the kind of thing a sanitizer job should see before merge.

### On the rest of #1081

Checked commit by commit against current main:

- **syncIdx out-of-bounds heap read** (the oldrel-1 segfault) — already in main; `git cherry` only flags it because context shifted.
- **#1011 dynamic Cmax/Cmin and #1014 sim.id/id** — reverted on the branch itself by `318022c97`; absent from its own tip.
- **`model(-a ~ NULL)` line removal** — superseded. Main has handled `~ NULL` drops since 2023-12-05 (`5ea14f94c`) and covers all three forms (`<-`, `=`, `~`) where the branch covers two; merging it would narrow main.
- **`odeMethodToInt()` multi-element method** — already in main, independently implemented with the same rationale (legacy nonmem2rx/monolix2rx callers passing the historical 4-element default).
- **Tutorial/learnr updates** — not carried here.

So #1081 can be closed once this lands.


mattfidler:
This didn't pass when I closed it.  I need to really carefully think about this. In theory this would be a good win for how many people use rxode2, but I want to be careful.  I'm unsure if altrep is actually saving time (sometimes it could take more).

PR #1385, opened 2026-09-19
Rebuild objects when a header changes

mattfidler:
`src/Makevars.in` declares no header dependencies, so `make` never rebuilds
anything when a header is edited.

Measured on a clean checkout of `main`, 267 objects built:

| touched | objects that rebuild |
| --- | --- |
| `src/<any>.h` | **0** |
| `inst/include/*.h` | **0** |

`make -q <objects>` returns 0 — "nothing to do" — in both cases.

The stale object then links against the previous signature. Sometimes that
fails loudly at load time with an undefined symbol; more often it silently
keeps running the **old** code long after a build that reported success. That
is a whole class of phantom test failure — results that do not match the source
in the tree — and it is expensive precisely because nothing warns you. The
usual conclusion drawn from one ("this branch broke those tests") is wrong, and
proving it wrong costs hours.

With the rule, the same touch rebuilds all 267 (179 `g++`, 81 `cc`, 7 Fortran)
and `make -q` reports out-of-date.

### Why `all: $(SHLIB)` is part of the fix

This is the reason it is not a one-line change, and it is easy to get wrong.

R **prepends** this file to its own makefiles — `tools:::.shlib_internal` does
`makefiles <- c("Makevars", makefiles)` — so the first target *here* becomes
the default goal for the entire build. Without `all: $(SHLIB)` ahead of it,
`$(OBJECTS)` becomes that goal: every build compiles the objects, stops, and
produces **no shared object at all**. `shlib.mk` has its own `all:`, but it is
in a later makefile and so cannot claim the default goal.

(For comparison, `nlmixr2est`'s `Makevars.in` already declares `all: $(SHLIB)`,
so the equivalent fix there is just the two `HEADERS` lines.)

### Verified end to end

After the change: a full `R CMD INSTALL` produces `rxode2.so`, `make` still
resolves to `all` ("Nothing to be done for 'all'"), and the installed package
loads and solves a model.

### Cost

One extra full rebuild whenever a shared header actually changes, which is the
correct behavior. A no-op install still rebuilds nothing. The rule is
deliberately coarse — every object depends on every header — matching the
approach that keeps it maintenance-free as sources come and go.


mattfidler:
@billdenney different solution to #1356 that doesn't add files CRAN doesn't want.

billdenney:
When I saw the description, I came to look because I was curious.

